PR #207, opened 2018-11-14
Manage dependencies with "go mod"

### LINKED ISSUE #203: Use vgo vendoring to deal with changing dependencies
Playing around further (after #202) I faced the following during `go build`:
```
# github.com/goki/gi/histyle
..\..\..\go\pkg\mod\github.com\goki\gi@v0.0.0-20181112192923-6c609aa46a98\histyle\hitags.go:288:28: undefined: chroma.EOFType
```
Looks like they changed the types in chroma a few days ago. Maybe its a good idea for gi to leverage modules with specific version dependencies.

xyproto:
* Add `go.mod` and `go.sum`, as generated by `go mod init`.
* Should resolve #203.
* Vendor dependencies with `go mod vendor`.
* Small adjustment to the import path in `gi/spell.go`, for importing `mvdan.cc/xurls/v2`.
* Using the latest commit of  `github.com/alecthomas/chroma` to resolve the `undefined: chroma.EOFType` issue.


xyproto:
All tests pass when running `go test`  in the `gi` directory.

komuw:
this project was not using the `vendor` directory before; maybe it should stay that way.

but the `go.mod` changes are really cool

xyproto:
The advantage of the vendor directory is that:

* This project can be built in the future, regardless of what happens to the various git repositories that it depends on.
* It's easier and cleaner for Linux distros to package, since the download and build phases can be cleanly separated, and the full project download is easier to checksum than checksumming every package that go get feels like downloading.

rcoreilly:
thanks!  I just read up on the modules stuff, and added a version v0.9.0 to the gi package, and a 'make version' target that updates the version info.  whenever we release a new version we need to remember to update the git tag and do that make command.  @rohrlich looping John in..

rcoreilly:
btw, I just removed the "vendor" part of it as it breaks the standard "go build" for gide, and seems like it is an additional step beyond the basic go.mod support -- I don't think we need it at this point, at least until I understand all this stuff better.

PR #418, opened 2020-11-27
add option to supress character echo in TextField

buck54321:
Setting `(TextField).NoEcho = true` will replace displayed characters with bullet points, as would be common in a password field.


rcoreilly:
Looks good -- thanks for the contribution!

PR #420, opened 2020-11-28
enable event propagation when Window not focused

buck54321:
I'm not sure if this is the right way to handle this. It would be incredibly useful if I could still leverage the `CustomEvent` system for synchronization and UI updates even when the `Window` is not focused. Preventing these `Event`s outright is perhaps overzealous, since the `CustomEvent` system seems like the natural place to synchronize UI updates across threads, and we can't control when the user will remove focus. This is particularly problematic on multi-monitor systems, since even though the `Window` isn't focused, it might still be fully visible. 

I think what you actually want to do is prevent **just the render** when the `Window` is **minimized**, but then apply the render when the `Window` is opened again. This would enable calling methods like `(*TextField).SetText` while the window is out of focus, but avoiding the overhead of a render until necessary.

### DIFF
--- a/gi/window.go
+++ b/gi/window.go
@@ -155,6 +155,7 @@ type Window struct {
 	PopupFocus        ki.Ki             `json:"-" xml:"-" desc:"node to focus on when next popup is activated -- use SetNextPopup"`
 	DelPopup          ki.Ki             `json:"-" xml:"-" desc:"this popup will be popped at the end of the current event cycle -- use SetDelPopup"`
 	PopMu             sync.RWMutex      `json:"-" xml:"-" view:"-" desc:"read-write mutex that protects popup updating and access"`
+	BlurEvents        bool              `json:"-" xml:"-" view:"-" desc:"propagate events when the window doesn't have focus"`
 	lastWinMenuUpdate time.Time
 	// below are internal vars used during the event loop
 	delPop        bool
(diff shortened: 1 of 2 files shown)


rcoreilly:
Several issues here:

* Rendering checks *visible* not focus, so it should happen outside of focus for sure.  Does SetText() not update? I'm surprised if it doesn't -- please file an issue for that if so.

* Rendering updates are also independent of the window event mechanism in general.

* Events are really mostly for oswin / glfw GUI-generated events.  I'm not even sure we *get* many / any such events outside of focus, so that would be something to look into first -- do we even need to check that?  I can't remember that far back in terms of whether that was necessary for something..

* If we do actually need to filter some events on focus, it might be better to use a an event interface method instead?  "Blur" might be a bit too clever (was for me for a sec there :) -- just `OnWinFocus()` would be clear and parallel to existing `OnFocus()`.  Then events that make sense to be filtered can be, and others can just pass through.  CustomEvent woudl definitely be

buck54321:
> Rendering checks visible not focus, so it should happen outside of focus for sure. Does SetText() not update? I'm surprised if it doesn't -- please file an issue for that if so.

I guess I just meant in the sense that a `CustomEvent` will likely entail some UI updates, which will schedule a render. But I didn't realize that you already blocked render on non-visibility, so my suggestion is moot. 

> If we do actually need to filter some events on focus, it might be better to use a an event interface method instead?

I tried that with 6312547. I have no clue whether any of the other concrete `Event` implementations should return `false`, but I'm happy make it so if you can think of some.

rcoreilly:
looks good -- can test later for other cases that might want to return false, but this will be safe for now.

rcoreilly:
I'll hold off making a new minor release in case you have some other changes, and after I fix the align in next couple of days, unless having such a release would be particularly useful for you at this point.

buck54321:
No need to rush. Thanks for asking.

PR #482, opened 2022-09-29
update vulkan dep

garymm:


rcoreilly:
btw I have one other thing to do in vulkan that I will do sometime soon hopefully, and then I will push a new release.

PR #746, opened 2024-01-03
replace all MSec with time.Duration

khalil-farashiani:
according to the https://github.com/goki/gi/issues/678 
I replace all MSec with time.Duration https://github.com/goki/gi/issues/678


kkoreilly (approved):

kkoreilly:
Thanks!

PR #801, opened 2024-01-14
test drop file

ddkwork:


rcoreilly:
That is a good idea to have it open directly.  We would want to go through the Drop event instead of directly accessing glfw, which is not used on all platforms.

ddkwork:
> That is a good idea to have it open directly.  We would want to go through the Drop event instead of directly accessing glfw, which is not used on all platforms.
> 

I deleted the previous way of forking the image repository to other websites, resetting the gitlab repository is too troublesome: you need the first step to sync github, the second step is to rebase to the last commit if you have a commit or lag behind your commit, in fact, it is to delete all my operations and keep it close to your last commit to solve the problem of too much difference when merging PR, The third step is to write PR code to push to gitlab, and at the same time automatically push to the github repository I fork through the gitlab image function, and the fourth step is to click the PR button... It's a hassle, and using a VPN saves half of the steps, but it's not free. The PR tests the drag-and-drop files and also tests whether the multi-person development is standardized. It's a hassle to use github fork directly for pr and sync in China, you have to use a vpn to do it, otherwise you can't access sites like github and Google, if you know of any code hosting platforms that are accessible in China and offer sync one-click reset and push mirroring for free, please let me know, I want to do this process in a simple way.

kkoreilly:
I do not know of any code hosting platforms that can do that, sorry. You don't have to go through the GitHub website for resetting your fork to the main branch; instead you can set up an upstream remote and fetch from that, but I'm not sure whether that would be any better, as you would still be going through the GitHub servers.

ddkwork:
> I do not know of any code hosting platforms that can do that, sorry. You don't have to go through the GitHub website for resetting your fork to the main branch; instead you can set up an upstream remote and fetch from that, but I'm not sure whether that would be any better, as you would still be going through the GitHub servers. 

Let's use the VPN method first, the other methods are too cumbersome

ddkwork:
> I do not know of any code hosting platforms that can do that, sorry. You don't have to go through the GitHub website for resetting your fork to the main branch; instead you can set up an upstream remote and fetch from that, but I'm not sure whether that would be any better, as you would still be going through the GitHub servers. 

The other way is for you to maintain a real-time synchronization of the GitHub-to-Gitlab repository, but this adds complexity to your work and is not worth it. It is accessible in China gitlab.com but not github.com and Google.com etc.

kkoreilly:
You shouldn't sync your fork when you have an active PR because it resets it. As long as you sync your fork every time before you start a PR, you should be fine and we shouldn't have merge conflicts.

PR #888, opened 2024-02-16
panic("reflect: slice index out of range")

ddkwork:


kkoreilly:
I figured out the cause of that problem, and that is not the best solution. I will work on implementing my fix.

kkoreilly:
I fixed this problem in 0d52caf18e4f15b0e1b54bc90cf6923a9aec1963.

ddkwork:
> I fixed this problem in 0d52caf18e4f15b0e1b54bc90cf6923a9aec1963.

At that time, I triggered this panic operation instead of double-clicking, but like this: fill the table with an empty struct, run the main function, right-click on the table layout, click edit, each row is empty, but we can come to the window where the struct is displayed, this process causes the index to be 0 because of the empty table, and the reflection is out of bounds

kkoreilly:
I fixed the other part of this issue that you mentioned. Now it both selects the correct row and is robust to right clicks out of bounds.

ddkwork:
thanks



---Original---
From: ***@***.***&gt;
Date: Sun, Feb 18, 2024 14:47 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [cogentcore/core] panic("reflect: slice index out of range") (PR#888)




 
I fixed the other part of this issue that you mentioned. Now it both selects the correct row and is robust to right clicks out of bounds.

Issue #917, opened 2024-02-28
treeview example : panic: reflect: Call using zero Value argument


c1ngular:
### Describe the bug

run treeview example , Stylers -> click "xx set styles" buttons inside the both popups , cause panic . 
<img width="431" alt="截屏2024-02-28 16 17 30" src="https://github.com/cogentcore/core/assets/9145844/7138f7d1-4a62-4b36-b791-c6d2663805e3">

### How to reproduce

`2024/02/28 16:05:48 panic: reflect: Call using zero Value argument
2024/02/28 16:05:48 
2024/02/28 16:05:48 ----- START OF STACK TRACE: -----
2024/02/28 16:05:48 goroutine 67 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x64
cogentcore.org/core/goosi.HandleRecoverBase({0x105230f60, 0x1400027d600})
	/Users/lijiangtv/go/src/core/goosi/recover.go:49 +0x38
cogentcore.org/core/gi.HandleRecover({0x105230f60, 0x1400027d600})
	/Users/lijiangtv/go/src/core/gi/recover.go:32 +0x48
cogentcore.org/core/gi.(*RenderWin).EventLoop.func1()
	/Users/lijiangtv/go/src/core/gi/renderwin.go:555 +0x34
panic({0x105230f60?, 0x1400027d600?})
	/usr/local/go/src/runtime/panic.go:914 +0x218
reflect.Value.call({0x1052348a0?, 0x140003d11e0?, 0x106ac8548?}, {0x104d85c25, 0x4}, {0x14000196168, 0x1, 0x0?})
	/usr/local/go/src/reflect/value.go:447 +0x15c4
reflect.Value.Call({0x1052348a0?, 0x140003d11e0?, 0x1400028d8e8?}, {0x14000196168?, 0x14000238880?, 0x105381298?})
	/usr/local/go/src/reflect/value.go:380 +0x94
cogentcore.org/core/giv.(*FuncButton).CallFuncShowReturns(0x1400838ca80)
	/Users/lijiangtv/go/src/core/giv/funcbutton.go:273 +0x190
cogentcore.org/core/giv.(*FuncButton).CallFunc.func2.1({0x105381298?, 0x1400ef3afc0?})
	/Users/lijiangtv/go/src/core/giv/funcbutton.go:314 +0x5c
[...]

Issue #916, opened 2024-02-28
 views example : panic: reflect: call of reflect.Value.IsZero on zero Value after clicking "2 table structs" button 


c1ngular:
### Describe the bug

panic raise in console and app freeze after after clicking "2 table structs" button  from views example  
<img width="273" alt="截屏2024-02-28 15 48 53" src="https://github.com/cogentcore/core/assets/9145844/b2f09bc3-7d40-4a6c-abea-9ffafe89fb64">

### How to reproduce

`2024/02/28 15:48:37 panic: reflect: call of reflect.Value.IsZero on zero Value
2024/02/28 15:48:37 
2024/02/28 15:48:37 ----- START OF STACK TRACE: -----
2024/02/28 15:48:37 goroutine 13 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x64
cogentcore.org/core/goosi.HandleRecoverBase({0x1010fc080, 0x1400009c720})
	/Users/lijiangtv/go/src/core/goosi/recover.go:49 +0x38
cogentcore.org/core/gi.HandleRecover({0x1010fc080, 0x1400009c720})
	/Users/lijiangtv/go/src/core/gi/recover.go:32 +0x48
cogentcore.org/core/gi.(*RenderWin).EventLoop.func1()
	/Users/lijiangtv/go/src/core/gi/renderwin.go:555 +0x34
panic({0x1010fc080?, 0x1400009c720?})
	/usr/local/go/src/runtime/panic.go:914 +0x218
reflect.Value.IsZero({0x0?, 0x0?, 0x14009e42780?})
	/usr/local/go/src/reflect/value.go:1638 +0x524
cogentcore.org/core/laser.OnePtrUnderlyingValue({0x1010a8860?, 0x14000349cd0?, 0x14004d131d0?})
	/Users/lijiangtv/go/src/core/laser/ptrs.go:102 +0x4c
cogentcore.org/core/giv.(*TableView).ConfigRows(0x14009f00000)
	/Users/lijiangtv/go/src/core/giv/tableview.go:357 +0x344
cogentcore.org/core/giv.(*TableView).ConfigTableView(0x14009f00000)
	/Users/lijiangtv/go/src/core/giv/tableview.go:238 +0xe8
cogentcore.org/core/giv.(*TableView).ConfigWidget(0x14009f00000?)
	/Users/lijiangtv/go/src/core/giv/tableview.go:227 +0x1c
[...]

Issue #679, opened 2023-12-25
Resolve handling of nil values


kkoreilly:
We need to determine a good way for nil values to be displayed and edited. This is related to #539 and #575.

kkoreilly:
We also need to fix panics after adding a nil function in a slice view (eg: when adding a function to Stylers). You probably just shouldn't be able to add new elements to slices, maps, and arrays of functions.

PR #918, opened 2024-02-28
fix https://github.com/cogentcore/core/issues/917 and #916

ddkwork:


ddkwork:
well done, please check pr.

ddkwork:
see here 
```go
// IsValid reports whether v represents a value.
// It returns false if v is the zero Value.
// If IsValid returns false, all other methods except String panic.
// Most functions and methods never return an invalid Value.
// If one does, its documentation states the conditions explicitly.
func (v Value) IsValid() bool {
	return v.flag != 0
}

// IsZero reports whether v is the zero value for its type.
// It panics if the argument is invalid.                                  ----------------> see here
func (v Value) IsZero() bool {
```

In reflection code, I suggest calling IsValid before calling IsZero, because only with a value can we determine whether the value is zero. For example, for a variable, define it first, and if it is defined, IsValid is OK. Then we can determine whether the variable has been initialized or assigned, and only then can we call IsZero, because reflection allows variables to be undefined beforehand. This feels a bit like the saying "chicken lays eggs, egg lays chicken", where there are chickens first and then eggs. If there are chickens first, IsValid is okay. At this point, if the chickens do not lay eggs

kkoreilly:
I am planning to fix these issues by implementing some solutions based on your ideas, but your ideas by themselves do not fully fix the problems. Thanks for your work on this!

ddkwork:
okay



---Original---
From: ***@***.***&gt;
Date: Thu, Feb 29, 2024 12:11 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [cogentcore/core] fixhttps://github.com/cogentcore/core/issues/917 and #916 (PR #918)




 
I am planning to fix these issues by implementing some solutions based on your ideas, but your ideas by themselves do not fully fix the problems. Thanks for your work on this!

ddkwork:
> I am planning to fix these issues by implementing some solutions based on your ideas, but your ideas by themselves do not fully fix the problems. Thanks for your work on this!

I summarized the problem I have been mentioning in the discussion forum, n more panic is the process like this: the example directory constructs the structure pointer variable is passed to the tree and table and other widgets, because the demo we don't care about instantiating the value check, resulting in a series of subsequent panics, which also verifies what you said earlier "the interface asserts that panic never happens", why is this happening? Because in real projects, we will be very strict to check whether all kinds of things instantiate each parameter or field, slice members, etc., because there is the correct value filling, it naturally can't panic. Guess the demo data transfer check is not strict, when it comes to ki to reflect it, part of the code implementation of ki needs new a reflected value and then fill in the value from the demo, at this time the demo parameter is nil, then the problem seen in the two issues has occurred, ctx is nil, call iszero, but it isvalid has not passed, the slice is out of bounds is similar, ignore the interface assertion I feel is the culprit, In the state where the demo parameters mentioned above are like that, it is difficult to assert that there is no panic. Then we all know how to avoid it, but in the real project, there are still some parameters that should not be inexplicably entered, such as the packet capture program, when the packet capture program is run for the first time, the agent is not set up, the table is empty at this time, if the hand is cheap at this time to click on the table header sorting, panic will happen, in addition to this kind of example, there are many more, so my final suggestion is: rectify the interface assertion.

kkoreilly:
Again, it is much nicer to fix this at the high level instead of adding nil checks to everything. I will continue to work on fixing the crashes in the context of #916, #917, and #679, and if I can not fix them at the high level, then I will follow your suggestions.

ddkwork:
okay



---Original---
From: ***@***.***&gt;
Date: Thu, Feb 29, 2024 13:08 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [cogentcore/core] fixhttps://github.com/cogentcore/core/issues/917 and #916 (PR #918)




 
Again, it is much nicer to fix this at the high level instead of adding nil checks to everything. I will continue to work on fixing the crashes in the context of #916, #917, and #679, and if I can not fix them at the high level, then I will follow your suggestions.

ddkwork:
this can changed to assert.True()
Nice work!

kkoreilly:
Thanks! It can't be `assert.True`, since we sometimes want it to be false (that is why we have `test.ok` in that test).

ddkwork:
Okay



---Original---
From: ***@***.***&gt;
Date: Thu, Feb 29, 2024 14:06 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [cogentcore/core] fixhttps://github.com/cogentcore/core/issues/917 and #916 (PR #918)




 
Thanks! It can't be assert.True, since we sometimes want it to be false (that is why we have test.ok in that test).

ddkwork:
Run the treeview demo and click on a random + sign in the middle of the screen to appear a mysterious null pointer, which seems to be located in the update function, but it doesn't panic every time, which is a bit difficult to debug, but I'd love to see which field or variable becomes nil.

ddkwork:
Have you come up with a global fix for this panic?
{8BF307A7-5697-4316-9690-4DC299542B71}

ddkwork:
//if not this is fix code
```go`
func IsZero(v reflect.Value) bool {
	if v.IsValid() {
		return false
	}
	return v.IsZero()
}
````

kkoreilly:
I am continuing to think about a fix for those issues, and I will work on implementing it soon. I appreciate your suggestions, and I should have enough information to fix the issues.

ddkwork:
Ok

PR #982, opened 2024-06-17
enable build on termux

MatejMagat305:
here is ndk for termux: https://github.com/Lzhiyong/termux-ndk, I until august do diploma, but you can use (as patern): https://github.com/MatejMagat305/fyne-termux, if not, I will do it after my diploma

### DIFF
--- a/cmd/core/mobile/env.go
+++ b/cmd/core/mobile/env.go
@@ -452,6 +452,9 @@ func ArchNDK() string {
 				arch = "x86_64"
 				break
 			}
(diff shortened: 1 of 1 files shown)


kkoreilly:
Thanks! I will let you handle the other termux steps whenever you are ready.

MatejMagat305:
> Thanks! I will let you handle the other termux steps whenever you are ready.

it is already finish, I only will make bash script to set all enviroment variables ..., if termux already has NDK_HOME and others it will work ...

PR #983, opened 2024-06-17
re-set java ctx on resume

MatejMagat305:
it is "only" sugest to fix same problem as fyne ............., on resume java context do not work ..., also I dare unused function move to C code to avoid GO runtime, of course you can reject for example some clever  way, but this was my idea to fix ..., you can free change it ...


kkoreilly (approved):
LGTM; thanks!

PR #990, opened 2024-06-24
solution to many issues in rendering by explicitly marking container frames

rcoreilly:


kkoreilly (changes requested):
These changes are unacceptable. As I quite clearly expressed in #962, parts are an unnecessary construct that results in significant additional complexity and confusion for no reason. Moreover, your parent border radius paradigm is deeply hacky and substitutes fixing the underlying math with adding even further unnecessary additional complexity.

PR #1001, opened 2024-07-05
spell package significantly simplified and now uses a simple dictionary list so embed size is much smaller

rcoreilly:


kkoreilly:
This reduces binary size by 14.0 MB!

PR #1036, opened 2024-07-24
IsNotSaved working again

rcoreilly:
key for saving changed files and save all in code etc.

### DIFF
--- a/texteditor/buffer.go
+++ b/texteditor/buffer.go
@@ -84,13 +84,12 @@ type Buffer struct { //types:add
 	// autoSaving is used in atomically safe way to protect autosaving
 	autoSaving bool
 
-	// Changed indicates if the text has been Changed (edited) relative to the
-	// original, since last EditDone
-	Changed bool
-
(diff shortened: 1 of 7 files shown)


texteditor/buffer.go
```diff
@@ -84,13 +84,12 @@ type Buffer struct { //types:add
...
 	autoSaving bool
 
-	// Changed indicates if the text has been Changed (edited) relative to the
-	// original, since last EditDone
-	Changed bool
-
-	// NotSaved indicates if the text has been changed (edited) relative to the
-	// original, since last Save
-	NotSaved bool
+	// notSaved indicates if the text has been changed (edited) relative to the
```
kkoreilly, line 87: Why not just export NotSaved?

rcoreilly: see note elsewhere -- we might need a mutex to protect that field -- this makes that possible without api changes.

kkoreilly (approved):

PR #1040, opened 2024-07-24
fix switch svgs (again)

rcoreilly:
-790 is the magic number.


kkoreilly (changes requested):
Please fix the merge artifacts.

kkoreilly (approved):

PR #1043, opened 2024-07-24
add lf at end of bytes -- turns out it is a standard..

rcoreilly:
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline


kkoreilly (approved):

kkoreilly (changes requested):
CI failing.

kkoreilly (approved):

PR #1045, opened 2024-07-24
temporarily disable text editors on mobile due to stability issues

kkoreilly:


rcoreilly:
delete this now, right?

kkoreilly:
Again, we still need to fix the issue where you can't scroll past the text editor before we can get rid of this branch. We can delete the PR though.

PR #1050, opened 2024-07-25
support for TilePlus in Splits

rcoreilly:
TilePlus is logical missing element in current splits tiles, and the most efficient use of tabs space in Cogent Code.


docs/content/2-widgets/2-containers/splits.md
```diff
@@ -37,6 +37,16 @@ core.NewText(sp).SetText("Fourth")
 sp.SetTiles(core.TileSpan, core.TileSecondLong)
 ```
 
+```Go
```
kkoreilly: You should add text here that says something like `There are various tile options for splits:`

kkoreilly (changes requested):

kkoreilly (approved):

PR #1053, opened 2024-07-25
Scroll adjusts scroll delta to match actual (on laptop anyway)

rcoreilly:
@kkoreilly can you test on iphone?  I'll test on android when I get a sec.


rcoreilly:
scroll looks good on android native.  checking android web.  safari web on macbook is too fast by maybe 2x, so we may need some platform specific factors.

rcoreilly:
android web is good.

kkoreilly:
My understanding of the issue is that the scrolling is too choppy and slow, not too fast relative to the actual movement. I encourage you to try scrolling the blog post on github and then on our website, and you will immediately notice the difference. The key features of the normal scrolling are:

1. It isn't choppy
2. It continues scrolling even after you let go
3. It both accelerates and decelerates, creating a very smooth experience

For us, the first is a performance issue that seems to be especially exacerbated by text editors and images. The second and third are features that we need to implement. This PR addresses none of them. I suggest that we close this PR and create an issue with this information instead.

kkoreilly:
Note that those second and third points seem to already happen with real scrolling (the system handles it for us), so we need to figure out a way to implement that algorithm in Go for slide scrolling.

kkoreilly:
See #1056 and #1058.

rcoreilly:
If you look at the comments, and do manual press-and-slide scrolling on any platform, you'll see that the rate of mouse movement is much slower than the rate of scrolling.  This very simple PR fixes that on most platforms.  It is just one very simple part of the bigger picture.

kkoreilly:
The rate of mouse movement is not supposed to be the same as the rate of scrolling; it should be much slower. You should not have to drag your mouse across the entire app for it to move one page. Again, this is not at all what people are complaining about.

rcoreilly:
"The scroll speed is faster than my drag amount" https://www.reddit.com/r/golang/comments/1eb6cd8/comment/lercfux/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

"and the scrolling is broken, doesn't match scroll speed" https://www.reddit.com/r/golang/comments/1eb6cd8/comment/leuab30/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

"the scroll acceleration feels a bit unnatural," https://www.reddit.com/r/golang/comments/1eb6cd8/comment/ler8tvd/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

go to any website on your phone and use the press-slide scrolling -- it matches exactly what you drag.  that is what people expect.

kkoreilly:
I guess if you press and slide without any flick, it does match the movement amount. I am just used to always accelerating, which does involve it moving a lot more than you slide.

kkoreilly (approved):

Issue #1056, opened 2024-07-25
Implement hardware accelerated rendering with WebGPU/Vello


kkoreilly:
### Describe the feature

Currently, scrolling on web is very laggy and choppy. We need to improve this by profiling and improving the performance.

kkoreilly:
Here is some profiling data from scrolling on our initial release blog post in Chrome on macOS:

* 65.3% of time is spent on widget rendering (not involving system rendering, layout, styles, events, etc)
* 33.8% on `image/draw.Draw` (11.0% from `core.Image.Render`, 10.9% from `paint.Context.DrawBox`, 7.9% from `paint.Text.Render`, and 3.6% from `system.DrawerBase.Copy`)
* 28.7% on `paint.Context.DrawStandardBox` (15.8% from `core.Frame.Render`, 2.7% from `core.Text.Render`)
* 22.2% on `core.Scene.layoutScene`
* 21.3% on `core.Image.Render`
* 16.3% on `paint.Text.LayoutStdLR` (11.0% from `core.Text.configTextSize` and 5.0% from `core.Text.configTextAlloc`)
* 16.2% on `core.Text.RenderWidget` (15.8% on `core.Text.Render`)
* 14.3% on `paint.Text.Render` (13.0% from `core.Text.Render`)
* 14.1% on `raster.Path.AddTo` (11.1% from `paint.Context.StrokePreserve` and 2.9% from `paint.Context.FillPreserve`)

Overall, this paints a very clear picture: the biggest bottleneck when scrolling a blog post on web is rendering, with text and images being major bottlenecks. Standard box rendering, mainly from frames and other assorted widgets, also contributes a lot.

rcoreilly:
it should not be calling layoutScene or LayoutStdLR during scrolling.  are you sure you limited the profiling to after the initial render and just when scrolling?  otherwise, it is just a lot of rendering, which is kind of inevitable..  you can also get this data on native mac -- likely the same right?

kkoreilly:
By use of print statements, I can definitively affirm that LayoutStdLR is happening during scrolling.

Different things may be more of a bottleneck on web versus other platforms; it is also good to confirm that the actual web rendering is only around 4% of the total time.

kkoreilly:
An actual `WidgetBase.NeedsLayout` is not happening however.

rcoreilly:
that would take 0% time and likely doesn't show up.  it really shouldn't be doing layoutscene otherwise.

[...]

PR #1059, opened 2024-07-26
use Restyle instead of Update in scroll position -- fixes excessive layout with texteditors

rcoreilly:
for #1056 -- tested on android device native and it seemed smoother, though still not great.

@kkoreilly can you test on various web platforms?  it is too slow here at the hospital.


kkoreilly:
This is much much better on all web platforms!

kkoreilly (approved):

PR #1060, opened 2024-07-26
switch to NearestNeighbor image resampling for styles.Style.ResizeImage; much better performance on web

kkoreilly:


kkoreilly:
This completely removes image lag on web without any significant deleterious impact on image quality.

rcoreilly:
nice!!  how'd you figure that out!?

rcoreilly (approved):

kkoreilly:
The image lag was only when it first rendered the image for a certain size, so I knew that it was due to the resizing, and NearestNeighbor is a much faster algorithm than Linear.

PR #1065, opened 2024-07-26
Reset key mods on blur in web

### LINKED ISSUE #1061: Key modifiers on web get in a corrupted state
### Describe the bug

Certain keyboard shortcuts on web result in the key modifiers not getting reset and thus ending up in a corrupted state.

### How to reproduce

Press `Cmd+Alt+J`/`Ctrl+Alt+J`, refocus the app, and then try to press other keys and observe that they do not work.

### Example code

_No response_

### Relevant output

_No response_

### Platform

Web

kkoreilly:
We need to reset the key mods when the app loses focus to prevent them from getting in an inaccurate state.

Fixes #1061


rcoreilly (approved):
blur is the worst name for losing focus!

kkoreilly:
JS loves it!

PR #1067, opened 2024-07-26
exclude icons from list (was already in table) and use consistent expression to exclude

### LINKED ISSUE #1066: selecting an icon in Settings/Favorite paths is crazy wide
### Describe the bug

see title

### How to reproduce

open Settings/System/fav paths -- click on an icon

### Example code

_No response_

### Relevant output

_No response_

### Platform

macOS

rcoreilly:
Fixes #1066

### DIFF
--- a/core/list.go
+++ b/core/list.go
@@ -486,7 +486,7 @@ func (lb *ListBase) BindSelect(val *int) *ListBase {
 func (lb *ListBase) UpdateMaxWidths() {
 	lb.maxWidth = 0
 	npv := reflectx.NonPointerValue(lb.elementValue)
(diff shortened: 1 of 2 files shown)


kkoreilly (changes requested):
Please use `reflect.TypeFor`.

kkoreilly (changes requested):
Good job using `reflect.TypeFor`; you should also make an `iconType` variable to make this code more DRY and avoid the cost of computing it every time.

rcoreilly:
not worth it.  this is called infrequently.

kkoreilly (approved):

PR #1068, opened 2024-07-26
Make slide scrolling smoother by adding a continued scrolling velocity

### LINKED ISSUE #1058: Make slide scrolling smoother
### Describe the feature

We need to add support for acceleration and deceleration to slide scrolling, in addition to making it so that it continues to scroll after you stop sliding.

### Relevant code

_No response_

kkoreilly:
If we have enough velocity, we continue scrolling over the next second in a goroutine while slowly decelerating for a smoother experience.

Fixes #1058

### DIFF
--- a/core/frame.go
+++ b/core/frame.go
@@ -5,7 +5,6 @@
 package core
 
 import (
-	"fmt"
 	"log/slog"
 	"time"
 	"unicode"
@@ -152,13 +151,27 @@ func (fr *Frame) Init() {
 		fr.scrollDelta(events.NewScroll(e.WindowPos(), del, e.Modifiers()))
 	})
 	fr.On(events.SlideStop, func(e events.Event) {
-		vel := math32.Vector2FromPoint(e.StartDelta()).DivScalar(float32(e.SinceStart().Milliseconds()))
-		hasX := math32.Abs(vel.X) > 0.8
-		hasY := math32.Abs(vel.Y) > 0.8
+		vel := math32.Vector2FromPoint(e.StartDelta()).DivScalar(float32(e.SinceStart().Milliseconds())).Negate()
+		hasX := math32.Abs(vel.X) > 1
+		hasY := math32.Abs(vel.Y) > 1
 		if !hasX && !hasY {
 			return
 		}
-		fmt.Println(vel)
+		// If we have enough velocity, we continue scrolling over the
+		// next second in a goroutine while slowly decelerating.
+		go func() {
+			i := 0
+			tick := time.NewTicker(time.Second / 60)
+			for range tick.C {
+				fr.scrollDelta(events.NewScroll(e.WindowPos(), vel, e.Modifiers()))
+				vel.SetMulScalar(0.9)
+				i++
+				if i > 60 {
+					tick.Stop()
+					break
+				}
+			}
+		}()
 	})
 }
 
--- a/core/frame.go
+++ b/core/frame.go
@@ -152,7 +152,8 @@ func (fr *Frame) Init() {
 		fr.scrollDelta(events.NewScroll(e.WindowPos(), del, e.Modifiers()))
 	})
 	fr.On(events.SlideStop, func(e events.Event) {
-		fmt.Println(e.StartDelta(), e.SinceStart())
+		vel := math32.Vector2FromPoint(e.StartDelta()).DivScalar(float32(e.SinceStart().Milliseconds()))
+		fmt.Println(vel)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


kkoreilly:
@rcoreilly can you test this on Android (web and/or native)? Also, do you think we need to call AsyncLock and AsyncUnlock around the scrolling? It seems to work both ways, and I don't think there should be any issues, but it might make sense to just be safe? Might that also have a negative performance impact though?

core/frame.go
```diff
@@ -144,12 +144,34 @@ func (fr *Frame) Init() {
...
+		// smoother experience.
+		vel := math32.Vector2FromPoint(e.StartDelta()).DivScalar(float32(e.SinceStart().Milliseconds())).Negate()
+		if math32.Abs(vel.X) < 1 && math32.Abs(vel.Y) < 1 {
+			return
+		}
+		go func() {
+			i := 0
+			tick := time.NewTicker(time.Second / 60)
+			for range tick.C {
+				fr.scrollDelta(events.NewScroll(e.WindowPos(), vel, e.Modifiers()))
```
rcoreilly, line 166: absolutely need AsyncLock / Unlock here.  time cost is minimal.

core/frame.go
```diff
@@ -144,12 +144,34 @@ func (fr *Frame) Init() {
...
+				vel.SetMulScalar(0.9)
```
rcoreilly: 0.9 is pretty rapid decay.. probably should at least have a gobal unexported for this.

events/source.go
```diff
@@ -94,6 +94,12 @@ func (es *Source) KeyChord(rn rune, code key.Codes, mods key.Modifiers) {
 // MouseButton creates and sends a mouse button event with given values
 func (es *Source) MouseButton(typ Types, but Buttons, where image.Point, mods key.Modifiers) {
 	ev := NewMouse(typ, but, where, mods)
+	if es.Last.MouseButtonType == MouseDown {
```
rcoreilly: shouldn't this be conditional on Up?  you definitely don't want to just have all down events be the same!?

kkoreilly: What do you mean? How does this make them all the same?

core/frame.go
kkoreilly: Done.

core/frame.go
kkoreilly: No point in having a global unexported if you don't need it. Don't add premature complexity! We can add a setting later if we want to. I will experiment with other values.

events/source.go
rcoreilly: You are copying from last always.

kkoreilly: I am just copying the start and prev info, which would otherwise be unset.

rcoreilly: Down should not have that info - only up.

kkoreilly: Done.

rcoreilly:
Last one resolved prematurely on my phone... Just exclude down for current when setting this info from last.

kkoreilly:
I already did.

core/frame.go
kkoreilly: What do you think about the speed now?

events/source.go
```diff
@@ -94,6 +94,12 @@ func (es *Source) KeyChord(rn rune, code key.Codes, mods key.Modifiers) {
 // MouseButton creates and sends a mouse button event with given values
 func (es *Source) MouseButton(typ Types, but Buttons, where image.Point, mods key.Modifiers) {
 	ev := NewMouse(typ, but, where, mods)
+	if typ == MouseUp && es.Last.MouseButtonType == MouseDown {
```
rcoreilly: sorry to be a PITA but the key thing is to just *exclude* MouseDown -- MouseDrag definitely needs these things.

rcoreilly: or maybe that goes in a different place?  anyway, just exclude MouseDown to be clearer.

kkoreilly: You know what the great thing about MouseDrag is? It goes through MouseMove, not MouseButton. The only things that go through MouseButton are MouseDown and MouseUp, but if you really want me to invert the logic for you, I can.

kkoreilly: Good?

kkoreilly:
The main remaining issue is that you can't get it to velocity scroll with a very small swipe due to the velocity threshold, but if you lower the threshold, then non-swipes incorrectly result in velocity scrolling.

rcoreilly:
do you want to play with that more or should we merge this?

kkoreilly:
Can you test it on Android and give feedback on speed and threshold?

kkoreilly:
We can figure out the speed and threshold later; this is already a major improvement. @rcoreilly please merge.

rcoreilly (approved):

Issue #660, opened 2023-12-25
Compute effective child bounds based on parent border radius to prevent children from rendering over their parents


kkoreilly:

kkoreilly:
I finally fixed this issue with a render-based fix in the standard box model that constraints the rendering bounds of elements based on the border radiuses of their parents. The various parts of the fix are contained in the commits above and other commits around them.

PR #1078, opened 2024-07-29
Improve table header styling

### LINKED ISSUE #1077: Table column headers are misaligned
### Describe the bug

There are a number of cosmetic issues with the table column headers,

- When a table is created the column headers are shifted slightly to the right. They are not correctly aligned with the table columns.
- When the first column header is clicked to enable sorting the header is expanded to show the direction of sort graphic, this in turn expands the whole column horizontally which is annoying. It would ideal to pre-allocate space for the sorting icon so the column header and its text doesn't jump about.

### How to reproduce

Create a simple table and click on the first column header.

### Example code

_No response_

### Relevant output

_No response_

### Platform

macOS

kkoreilly:
Fixes #1077.

### DIFF
--- a/core/toolbar.go
+++ b/core/toolbar.go
@@ -217,10 +217,10 @@ func ToolbarStyles(w Widget) {
 	wb.FinalStyler(func(s *styles.Style) {
 		if s.Direction == styles.Row {
 			s.Grow.Set(1, 0)
-			s.Padding.SetHorizontal(units.Dp(ConstantSpacing(16)))
+			s.Padding.SetHorizontal(units.Dp(16))
 		} else {
 			s.Grow.Set(0, 1)
-			s.Padding.SetVertical(units.Dp(ConstantSpacing(16)))
+			s.Padding.SetVertical(units.Dp(16))
 		}
 	})
 	wb.SetOnChildAdded(func(n tree.Node) {
--- a/core/slider_test.go
+++ b/core/slider_test.go
@@ -156,7 +156,7 @@ func TestSliderInput(t *testing.T) {
 		}
 		sr.SystemEvents().MouseButton(events.MouseUp, events.Left, image.Pt(200, 20), 0)
 	}, func() {
-		tolassert.EqualTol(t, 4, float32(n), 3)
+		tolassert.EqualTol(t, 5, float32(n), 4)
 		tolassert.Equal(t, 0.5756579, value)
 	})
 }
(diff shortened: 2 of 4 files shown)


core/toolbar.go
```diff
@@ -217,10 +217,10 @@ func ToolbarStyles(w Widget) {
 	wb.FinalStyler(func(s *styles.Style) {
 		if s.Direction == styles.Row {
 			s.Grow.Set(1, 0)
-			s.Padding.SetHorizontal(units.Dp(ConstantSpacing(16)))
+			s.Padding.SetHorizontal(units.Dp(16))
```
rcoreilly, line 220: why is this in here and what is the motivation?

rcoreilly (commented):
can you also implement in the tensor tableview so I can see what it looks like at scale, on the plotview data?

also what is up with the test?  just that random thing?

core/toolbar.go
rcoreilly: If this is to improve alignment in the table, which it seems like from your comment, then it should be styled specifically in the table, not in the generic toolbar styles.  For those using compact spacing, it is essential to have this constant spacing for the general toolbar case.

kkoreilly:
Yes, I just updated `tensorcore.Table` to the new changes, and it looks much better to me on the plot example data. The test failure is one of our two random test failures; I will work on making that more robust now.

core/toolbar.go
kkoreilly: With our new border radius based parent border radius encroachment fix (#660), we don't need constant spacing on the toolbar anymore.

kkoreilly: The table *does* now have specific styling, which is zero padding, but again, independent of that the toolbar does not need to have the constant spacing anymore.

rcoreilly: ok, I tested it on various use-cases and toolbars look fine..

kkoreilly: So did I.

rcoreilly:
FYI the change results in 3.5 fewer columns of data visible in the plot data table.  also it seems like it is reserving too much extra space?  e.g., the Epoch column doesn't need that much?

New:

image
 
Existing:

image

rcoreilly:
Indicator styling definitely looks better though!

kkoreilly:
You are certainly free to investigate the space allocation!

kkoreilly:
Also @rcoreilly do you have any interest in fixing this test timeout issue? You can look at the CI logs.

rcoreilly:
no interest in that.  should I merge anyway?

rcoreilly (approved):
Filed #1080 to deal with optimizing spacing, later..

kkoreilly:
Yes, merge, since CI shouldn't fail on main. We should definitely fix this though.

PR #1090, opened 2024-07-30
get rid of any validation tooltip in TextField.clearError

### LINKED ISSUE #1085: Text Input Validator Tooltip Misbehaves with Keyboard Commit
### Describe the bug

A text field with a validator defined permanently shows the tooltip with the validation error when you commit a value by hitting ctrl/command+enter.  Resolving the validation error and committing the new text does not clear the tooltip.

However, mousing over the text field will clear the tooltip.

### How to reproduce

1. Go to the documentation page for text fields: https://www.cogentcore.org/core/widgets/inputs/text-fields
2. Enter something like "Test" in the example text field that specifies a Validator to ensure the string contains "Go"
3. Ensure the mouse cursor is _NOT_  hovering the text field
4. Hit ctrl+enter (or command+enter on Mac) to commit the text that will fail validation
5. Observe the validation error tooltip with "Must contain Go" is displayed as expected, however it never goes away
6. Update the text to "Test Go" so that it no longer fails validation and hit ctrl+enter (or command+enter on Mac) to commit it
7. Observe the validation error tooltip is still visible.
8. Move the mouse cursor over the text field and observe the tooltip is cleared

### Example code

```go
tf := core.NewTextField(b)
tf.SetValidator(func() error {
	if !strings.Contains(tf.Text(), "Go") {
		return errors.New("Must contain Go")
	}
	return nil
})
```

### Relevant output

_No response_

### Platform

Web

kkoreilly:
Fixes #1085

### DIFF
--- a/core/textfield.go
+++ b/core/textfield.go
@@ -555,6 +555,7 @@ func (tf *TextField) clearError() {
 	}
 	tf.error = nil
 	tf.Update()
+	tf.Send(events.LongHoverEnd) // get rid of any validation tooltip
 }
 
 // validate runs [TextField.Validator] and takes any necessary actions


rcoreilly (approved):
Nice -- this will fix the app chooser too presumably.

PR #1092, opened 2024-07-30
send LongPressEnd in cancelLongPress when appropriate

### LINKED ISSUE #1086: Long press tooltip on slider does not update or disappear on value change
### Describe the bug

Long pressing on a slider to show the tooltip with the current value and then dragging the slider to change its value does not update the tooltip or remove it resulting in incorrect value display.

### How to reproduce

1. Go to the documentation page for tooltips: https://www.cogentcore.org/core/widgets/basic/tooltips
2. Long press on the slider to cause the tooltip with its current value to display and keep the button depressed
3. Drag the slider to change its value
4. Observe that the tooltip neither updates nor disappears

### Example code

```go
core.NewSlider(b)
```

### Relevant output

_No response_

### Platform

Web

kkoreilly:
Fixes #1086

### DIFF
--- a/core/events.go
+++ b/core/events.go
@@ -482,11 +482,6 @@ func (em *Events) handlePosEvent(e events.Event) {
 		}
 		em.press = nil
 
-		// if we have sent a long press start event, we send an end
-		// event (non-nil widget plus nil timer means we already sent)
-		if em.longPressWidget != nil && em.longPressTimer == nil {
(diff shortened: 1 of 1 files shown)


rcoreilly (commented):
You sure you tested this thoroughly?  I worry about regressions in something so general.

kkoreilly:
Yes, I tested it and also looked at all of the places that it is called, and it seems good.

rcoreilly (approved):

Issue #1083, opened 2024-07-30
Improve screen-specific settings structure


kjk:
Repro (on latest Chrome, latest Windows):
* go to https://www.cogentcore.org/
* select `Settings`
* click `+` button next to `Screens`

First, it's completely mysterious what it does.

Second: it adds `Zoom` field with value `10` (10% ?). This makes display almost un-readable.

Changing the `Zoom` value doesn't work. There's no way to remove it. 

I had to delete website data in Chrome to get back to normal state.

rcoreilly:
good points -- this is intended to be used with the button in the toolbar "Save screen zoom" so you can have different zoom factors for different screens.  there are still some issues with this as well  https://github.com/cogentcore/core/issues/940 .  we'll update the docs on it at least.

kkoreilly:
Yes, that `Screens` field is not structured well right now. We will work on improving that.

kkoreilly:
I fixed the formatting so that it says 10% instead of 10. Also, the tooltip states that these are "screen-specific settings, which will override overall defaults if set." Finally, in #1093, I made it so that pressing `+` has no actual effect, since the one and only screen on single-window platforms is now called "main", not "". None of those things resolve the issue that it is still somewhat confusing, but at least now it should be less confusing and will not have any immediate deleterious effects. We will work on improving the structure of screen-specific settings in general later. Thank you for reporting this!

(In terms of improving the structure of screen-specific settings, we need to make them be applied better (also see #940), make the default zoom 100%, and make it easier to select the names of available screens for the screen-specific settings map keys).

kkoreilly:
As of #1341 this structure is much better, with the `Screens` field being read-only and only set through the `Save screen zoom` button. Zooming with keyboard shortcuts also now changes the screen zoom if there is one saved, so zoom keyboard shortcuts will always be effective.

PR #1093, opened 2024-07-30
make default screen name main for AppSingle so that + on screen settings does not immediately screw things up

kkoreilly:
This makes #1083 much less potent of an issue.


rcoreilly (approved):
nice simple fix.

kkoreilly:
CI keeps timing out; just merge?

Issue #1047, opened 2024-07-24
Fix index page generation for higher-level pages without an index


kkoreilly:
### Describe the bug

If you have `a/b.md`, it will make an `a/b/index.html`, but not an `a/index.html` in `core build web`.

### How to reproduce

See above.

### Platform

macOS

kkoreilly:
This is superseded by #1321.

PR #1100, opened 2024-08-01
Implement HTML preview generation and improve SEO

### LINKED ISSUE #702: Implement SEO on web
We need to figure out a good way to handle URLs on web and have good SEO for sites. #701 is related.

kkoreilly:
This PR adds a new `generatehtml` feature that converts a Cogent Core app to HTML. It is not guaranteed to be perfect HTML, and it should not be used as a replacement for a Cogent Core app. However, it is good enough to be used as a preview and for SEO purposes, which this PR does.

This has several positive side effects, including that we now automatically generate `meta` descriptions for pages from their first paragraph, which further improves SEO. Also, it allows you to omit the `content` pages directory setting in config files, since that is now the default for apps that import pages.

This PR also contains a web loading screen improvement that makes it so that the loading screen is displayed up until the moment that the canvas is actually rendered, removing an annoying blank screen during that intermediate period. This PR also contains various cleanup items such as unexporting internal logic.

This PR will also allow us to soon implement an automatic navigation footer that links all pages together, which will also improve SEO support. We can also now use the `PreRenderData` as the basis for which pages to include in the generated path tree, which will allow us to more easily implement a fix for #1047.

The only remaining major issue here is that this relies on `RunMainWindow` being called, which does not happen in some apps like Cogent Code. I will work on figuring out a workaround for that.

Fixes #702


kkoreilly:
The key things is that we need to intercept the `Body` and prevent the window from actually being run, so to avoid the Cogent Code problem with apps that don't call `RunMainWindow`, it probably just makes sense to move the logic to `RunWindow` and add a guard so it is only called once.

kkoreilly:
I redesigned the `ExternalParent` logic for `RunWindow` and `RunMainWindow` and everything is working now.

rcoreilly (approved):
Wow, nice work!  I 'm not giving this a deep review b/c it is too much stuff but I've seen the screenshots of it working :)

PR #1136, opened 2024-08-14
timeinput date or time only

### LINKED ISSUE #1120: Add a date input
### Describe the feature

Similar to `core.TimeInput` but only renders a date field.

I couldn't find one but managed to get something that seems to work as expected with this code (copied from `core.TimeInput`. Also adds the ability to format the date (with a default, not sure if it's the best/correct way to set the default format).

Happy to create a PR, would it go in `timepicker.go`?

### Relevant code

```go
func NewDateInput(parent ...tree.Node) *DateInput { return tree.New[DateInput](parent...) }

// DateInput presents a text field for editing a date,
// which can pull up a [core.DatePicker] dialog.
type DateInput struct {
	core.Frame
	Time   time.Time
	Format string
}

func (ti *DateInput) WidgetValue() any { return &ti.Time }

func (ti *DateInput) SetFormat(f string) *DateInput {
	ti.Format = f
	return ti
}

func (ti *DateInput) Init() {
	ti.Frame.Init()

	if ti.Format == "" {
		ti.Format = "1/2/2006"
	}

	style := func(s *styles.Style) {
		s.Min.X.Em(8)
		s.Max.X.Em(10)
		if ti.IsReadOnly() { // must inherit abilities when read only for table
			s.Abilities = ti.Styles.Abilities
		}
	}

	tree.AddChild(ti, func(w *core.TextField) {
		w.SetTooltip("The date")
		w.SetLeadingIcon(icons.CalendarToday, func(e events.Event) {
			d := core.NewBody().AddTitle("Select date")
			dp := core.NewDatePicker(d).SetTime(ti.Time)
			d.AddBottomBar(func(parent core.Widget) {
				d.AddCancel(parent)
				d.AddOK(parent).OnClick(func(e events.Event) {
					ti.Time = dp.Time
					ti.UpdateChange()
				})
			})
			d.RunDialog(w)
		})
		w.Styler(style)
		w.Updater(func() {
			w.SetReadOnly(ti.IsReadOnly())
			w.SetText(ti.Time.Format(ti.Format))
		})
		w.SetValidator(func() error {
			d, err := time.Parse(ti.Format, w.Text())
			if err != nil {
				return err
			}
			// new date and old time
			ti.Time = time.Date(d.Year(), d.Month(), d.Day(), ti.Time.Hour(), ti.Time.Minute(), ti.Time.Second(), ti.Time.Nanosecond(), ti.Time.Location())
			ti.SendChange()
			return nil
		})
	})
}
```

j0hnsmith:
Adds `DisplayDate()` and `DisplayTime()` methods to `timeinput` so that optionally only a date or only a time can be displayed/selected. Methods used to allow chaining, can export the struct fields and remove the methods if prefered. 

Also adds hiding the time/date via `display:date` & `display:time` struct tags.

Fixes #1120


core/timepicker.go
```diff
@@ -307,13 +308,44 @@ func (dp *DatePicker) Init() {
 type TimeInput struct {
 	Frame
 	Time time.Time
+
+	// whether date input is rendered
+	displayDate bool
```
kkoreilly: Yes, I would prefer if these are exported. Also, for the documentation, it would be better to say `DisplayDate is whether the date input is displayed (default true).` etc. After you export them, you can run `go generate` and then it will automatically generate the setters.

core/timepicker.go
```diff
@@ -307,13 +308,44 @@ func (dp *DatePicker) Init() {
...
+
+// DisplayTime can be used to hide the time field (both date and time displayed by default).
+func (ti *TimeInput) DisplayTime(v bool) *TimeInput {
+	ti.displayTime = v
+	return ti
+}
+
+func (ti *TimeInput) OnBind(value any, tags reflect.StructTag) {
+	val := tags.Get("display")
+	switch val {
```
kkoreilly: You can inline `val` here (`switch tags.Get("display") {`)

kkoreilly (changes requested):
Thanks for filing this PR! Overall it looks good except for my two specific comments.

kkoreilly (approved):
Great, thank you! (Also note that I would typically ask you to add tests, but there are none for the time input yet, so I will add them myself in a separate PR now).

Issue #1007, opened 2024-07-13
how to avoid duplicate names in tree updates?  currently easy to crash in docs tree examples.


rcoreilly:
### Describe the bug

need to make tree copy / paste / drag-n-drop more robust to duplicate names, which can be easily created in many cases.  don't want the app to crash in this case: need to rename the child.  need to decide at what level to do this etc.

### How to reproduce

in docs app, in Collections/Trees, drag first `node-base-0` onto `node-base-1` and do `Add To Children`..

### Relevant output

```shell
Crash log saved in /Users/oreilly/Library/Cogent\ Core\ Docs/crash-logs

Platform: MacOS
System platform: MacOS
App version: v0.2.0-408-ge38cea5a2
Core version: v0.2.0-408-ge38cea5a2
Time: 2024-07-13 00:13:59

panic: plan.Update: duplicate name: tv_node-base-0

goroutine 51 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go/1.22.1/libexec/src/runtime/debug/stack.go:24 +0x64
cogentcore.org/core/core.handleRecover({0x106ab5560, 0x14014c57c60})
	/Users/oreilly/go/src/cogentcore.org/core/core/recover.go:38 +0x74
cogentcore.org/core/core.(*renderWindow).eventLoop.func1()
	/Users/oreilly/go/src/cogentcore.org/core/core/renderwindow.go:393 +0x34
panic({0x106ab5560?, 0x14014c57c60?})
	/opt/homebrew/Cellar/go/1.22.1/libexec/src/runtime/panic.go:770 +0x124
cogentcore.org/core/base/plan.Update[...](0x14012179538, 0x2, 0x140112995b8, 0x14011299598, 0x14011299580, 0x106d61ed0)
	/Users/oreilly/go/src/cogentcore.org/core/base/plan/update.go:50 +0x864
cogentcore.org/core/tree.UpdateSlice(0x2?, {0x106d83cd8?, 0x14012179508?}, {0x14014c6bbc0?, 0x14015229b80?, 0xb?})
	/Users/oreilly/go/src/cogentcore.org/core/tree/typeplan.go:34 +0xac
cogentcore.org/core/tree.Update({0x106d83cd8, 0x14012179508}, {0x14014c6bbc0, 0x2, 0x2})
	/Users/oreilly/go/src/cogentcore.org/core/tree/typeplan.go:56 +0x58
[...]

PR #1137, opened 2024-08-14
Prevent duplicate tab names to allow multiple tabs with the same label

### LINKED ISSUE #1131: Tab names need to be unique
### Describe the feature

Currently, the name used for a tab must be unique, otherwise the selection of tabs with the same name is somewhat broken.
It would be useful to have the ability to name tabs with the same name. Alternatively, the documentation should be changed to emphasise that tab names are the identifier and must be unique.

### Relevant code

_No response_

kkoreilly:
Fixes #1131


rcoreilly (approved):
how about doing this for the core.tree duplicate names too?

kkoreilly:
That would be possible (#1007).

Issue #1142, opened 2024-08-17
Chooser not aligned with TextField


j0hnsmith:
### Describe the bug

When a `time.Duration` field is in a struct that's used in a table, the `Chooser` is taller than the `TextField` in the `Spinner` (for the value) which doesn't look great (maybe this should be an enhancement since it's only visually 'broken').

### How to reproduce

Run the code below.

### Example code

```go
type SomeStruct struct {
    Took time.Duration
}
core.NewTable(b).SetSlice(&[]SomeStruct{{time.Duration(time.Minute)}, {time.Duration(time.Hour})
```

### Platform

macOS

kkoreilly:
Thank you for reporting this. I will look into this and see what I can do.

Issue #677, opened 2023-12-25
Improve the styling and behavior of disabled and read only elements


kkoreilly:
Various elements are not listening to whether they are disabled and/or read only, and even if they are, they are not styling and behaving correctly. Also, these states are not being properly inherited by different elements. One part of this #551.

kkoreilly:
We also need to fix read only time values, especially in the file view.

kkoreilly:
I fixed read-only switch handling in #1495, but there is still more logic required, probably for widgets like switches (plural) and color pickers for example. We also need to document read-only and disabled better and also improve naming (#1145), which should happen before more fixing given the breaking changes that will entail.

PR #1143, opened 2024-08-17
Improve handling of read only duration input and chooser

### LINKED ISSUE #1141: time.Duration in a read only table isn't read only
### Describe the bug

When a struct has a `time.Duration` field and is used in a read only table, the `DurationInput` isn't read only.

### How to reproduce

Run this code.

### Example code

```go
type SomeStruct struct {
    Took time.Duration
}
core.NewTable(b).SetSlice(&[]SomeStruct{{time.Duration(time.Minute)}, {time.Duration(time.Hour}).SetReadOnly(true)
```

### Relevant output

_No response_

### Platform

macOS

kkoreilly:
Fixes #1141. More progress for #677.

Some issues remain for read only duration input styling, but those can be handled as part of #1142. One question is whether we want to make `ReadOnly` automatically inherited, which would make things like this easier. There is also an outstanding issue regarding events being captured by read only multi-child widgets in tables, which we can fix as part of #677.

### DIFF
--- a/core/chooser.go
+++ b/core/chooser.go
@@ -173,24 +173,28 @@ func (ch *Chooser) Init() {
 	ch.SetIcon(icons.None).SetIndicator(icons.KeyboardArrowDown)
 	ch.CurrentIndex = -1
 	ch.Styler(func(s *styles.Style) {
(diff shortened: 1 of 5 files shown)


rcoreilly (approved):
yeah definitely need a better soln than this!

PR #1147, opened 2024-08-18
updates to cosh and cli to support cogent author, including cosh transpile command

rcoreilly:


cli/config.go
```diff
@@ -15,14 +15,18 @@ import (
 	"cogentcore.org/core/base/logx"
 )
 
+// TheMetaConfig holds the current [MetaConfig] data for access
+// of verbose and quiet state from client apps.
+var TheMetaConfig *MetaConfig
```
kkoreilly: As I previously told you, you can obtain the log level through `logx.UserLevel`. Unexport everything and use that in Cogent Author where you need it.

shell/cmd/cosh/cosh.go
```diff
@@ -114,3 +117,22 @@ func Build(c *Config) error {
 	}
 	return shell.NewShell().TranspileFile(c.Input, c.Output)
 }
+
+// Transpile transpiles all .cosh files in current directory
+// into corresponding .go files, without any further additions.
+// Use this for a multi-file codebase that is only used as a
+// go compiled package, not in interactive mode.
+func Transpile(c *Config) error {
```
kkoreilly: It seems to me that you could very easily detect whether these `package main` etc wrappers are already present and then add them if necessary in one unified `Build` command. This distinction between `Build` and `Transpile` is insufficient and will almost certainly cause confusion later. We don't want this to be like PHP. Alternatively, if you really can't automatically do it, add a bool flag for the unified `Build` command that specifies the behavior. The names `Build` and `Transpile` have absolutely no meaningful distinction.

shell/shell.go
```diff
@@ -337,6 +346,27 @@ func (sh *Shell) TranspileFile(in string, out string) error {
 	return errors.Join(err, werr)
 }
 
+// TranspileOnly only does direct transpiling of input to output
+// without adding anything to the resulting file.
+func (sh *Shell) TranspileOnly(in string, out string) error {
```
kkoreilly: This, similar to what I said about `Build` and `Transpile`, should be one function that either determines the distinction automatically or takes a bool flag for it. It seems like it would be very easy to detect if there is a `package main` line in the source code.

kkoreilly (changes requested):
Please address my comments.

shell/cmd/cosh/cosh.go
rcoreilly: build assumes one file, transpile iterates over all files in the directory.

shell/cmd/cosh/test.cosh
```diff
@@ -1,6 +1,7 @@
 // test file for cosh cli
 
-echo {args}
+// todo: no longer works
+// echo {args}
```
kkoreilly: What about fixing this?

rcoreilly: I forgot what we did to args -- working on other things now but if you remember, could you perhaps just do it?

kkoreilly: We seem to handle it in the `Run` command, so I don't think it is necessarily supposed to work with `Build`:

```
	if len(c.Args) > 0 {
		in.Eval("args := cosh.StringsToAnys(" + fmt.Sprintf("%#v)", c.Args))
	}
```

kkoreilly (approved):

PR #1149, opened 2024-08-18
improve gpu selection logic and webgpuinfo output

rcoreilly:
previously was completely excluding gl instead of not preferring it.  decodes key enums on webgpuinfo.

### DIFF
--- a/gpu/cmd/webgpuinfo/README.md
+++ b/gpu/cmd/webgpuinfo/README.md
@@ -1,0 +1,8 @@
+# webgpuinfo
+
+This tool prints out information about your available WebGPU devices.  Do this:
+
+```sh
+go run cogentcore.org/core/gpu/cmd/webgpuinfo@latest
+```
+
--- a/gpu/cmd/webgpuinfo/README.md
+++ b/gpu/cmd/webgpuinfo/README.md
@@ -3,6 +3,6 @@
 This tool prints out information about your available WebGPU devices.  Do this:
 
 ```sh
-go run cogentcore.org/core/gpu/cmd/webgpuinfo@latest
+go run cogentcore.org/core/gpu/cmd/webgpuinfo@main
 ```
 
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -176,52 +176,35 @@ func (gp *GPU) SelectGPU(gpus []*wgpu.Adapter) int {
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


gpu/cmd/webgpuinfo/README.md
```diff
@@ -0,0 +1,8 @@
+# webgpuinfo
+
+This tool prints out information about your available WebGPU devices.  Do this:
```
kkoreilly, line 3: Instead of this README, add a package comment of the form `Command webgpuinfo prints out...`. In general, we prefer package comments over READMEs when there is not much information, and we *never* create both a README and a package comment with the exact same information.

gpu/gpu.go
```diff
@@ -176,52 +176,35 @@ func (gp *GPU) SelectGPU(gpus []*wgpu.Adapter) int {
...
-	// gp.DeviceName = devNm
-	// if Debug {
-	// 	log.Printf("vgpu: %d selected device named: %s, memory size: %d\n", maxIndex, devNm, maxSz)
-	// }
-	// return maxIndex
+	return best
+}
+
+func gpuIsGLdBackend(bet wgpu.BackendType) bool {
+	return bet == wgpu.BackendTypeOpenGL || bet == wgpu.BackendTypeOpenGLES
```
kkoreilly, line 203: And you tested these changes on the Linux and Windows devices? Are we sure this isn't going to break anything?

gpu/cmd/webgpuinfo/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+```sh
+go run cogentcore.org/core/gpu/cmd/webgpuinfo@latest
```
kkoreilly: As I said previously, `@latest` is rarely the version you actually want while developing, so either change this to `@main` or get rid of it; you don't need to clarify how to run a Go command.

kkoreilly (changes requested):
Please address my comments.

gpu/cmd/webgpuinfo/README.md
rcoreilly: I for one always browse things on github where you see the README and NOT the doc comment.

gpu/gpu.go
rcoreilly: yes fully tested.

kkoreilly (approved):

PR #1164, opened 2024-08-20
time.sleep needs to run even for invisible -- allows inner-loop …

rcoreilly:
updates to get more frequent sleeps.. this is what works for emer sims.


kkoreilly:
Won't this lead to excessive sleeping if you having multiple plots?

rcoreilly:
actually that was critical for it to work in the first place!  e.g., epoch plot updates are pretty infrequent, not enough..

kkoreilly (approved):

PR #1165, opened 2024-08-21
Modeless dialog substitute for NewWindow dialogs on large single-window platforms

rcoreilly:
If sizeclass is Expanded, single-window web apps support "NewWindow" desktop-like experience with modal dialogs.

Fixed the problem of events leaking through the modal dialog.  Need to be on the lookout for any regressions from these event changes -- will test more thoroughly.

### DIFF
--- a/core/events.go
+++ b/core/events.go
@@ -400,6 +400,7 @@ func (em *Events) handlePosEvent(e events.Event) {
 			}
 			em.dragHovers = em.updateHovers(hovs, em.dragHovers, e, events.DragEnter, events.DragLeave)
 			em.dragMove(e)                              // updates sprite position
+			em.drag.AsWidget().HandleEvent(e)           // raw drag
 			em.drag.AsWidget().Send(events.DragMove, e) // usually ignored
 			e.SetHandled()
 		} else {
--- a/core/events.go
+++ b/core/events.go
@@ -249,11 +249,13 @@ func (em *Events) handlePosEvent(e events.Event) {
 		em.resetOnMouseDown()
 	case events.MouseDrag:
 		if em.spriteSlide != nil {
+			e.SetHandled()
 			em.spriteSlide.handleEvent(e)
 			em.spriteSlide.send(events.SlideMove, e)
 			return
 		}
 		if em.slide != nil {
+			e.SetHandled()
 			em.slide.AsWidget().HandleEvent(e)
 			em.slide.AsWidget().Send(events.SlideMove, e)
 			return
@@ -397,17 +399,19 @@ func (em *Events) handlePosEvent(e events.Event) {
 				}
 			}
 			em.dragHovers = em.updateHovers(hovs, em.dragHovers, e, events.DragEnter, events.DragLeave)
-			em.dragMove(e)                              // updates sprite position
-			em.drag.AsWidget().HandleEvent(e)           // raw drag
+			em.dragMove(e) // updates sprite position
+			e.SetHandled()
 			em.drag.AsWidget().Send(events.DragMove, e) // usually ignored
 		} else {
 			if em.dragPress != nil && em.dragStartCheck(e, DeviceSettings.DragStartTime, DeviceSettings.DragStartDistance) {
 				em.cancelRepeatClick()
 				em.cancelLongPress()
+				e.SetHandled()
 				em.dragPress.AsWidget().Send(events.DragStart, e)
 			} else if em.slidePress != nil && em.dragStartCheck(e, DeviceSettings.SlideStartTime, DeviceSettings.DragStartDistance) {
 				em.cancelRepeatClick()
 				em.cancelLongPress()
+				e.SetHandled()
 				em.slide = em.slidePress
 				em.slide.AsWidget().Send(events.SlideStart, e)
 			}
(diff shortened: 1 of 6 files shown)


rcoreilly:
@kkoreilly how do we get tags here to be able to open new window dialog from struct tags:

core/value.go:118
```go
	wb.OnClick(func(e events.Event) {
		if allowReadOnly || !wb.IsReadOnly() {
			wb.setFlag(e.HasAnyModifier(key.Shift), widgetValueNewWindow)
			openValueDialog(v, make, after...)
		}
	})
```

also shouldn't the same Expanded size class logic apply to mobile platform too (i.e., ipad)?

core/events.go
```diff
@@ -249,11 +249,13 @@ func (em *Events) handlePosEvent(e events.Event) {
 		em.resetOnMouseDown()
 	case events.MouseDrag:
 		if em.spriteSlide != nil {
+			e.SetHandled()
```
kkoreilly: This seems like the wrong approach to me. Instead of setting everything to handled, why not just prevent propagation of events that get sent to modeless dialogs directly in the central processing logic? Setting everything to handled seems like it will have a lot of bad consequences.

core/mainstage.go
```diff
@@ -165,17 +165,25 @@ func (st *Stage) firstWindowStages() *stages {
 
 // configMainStage does main-stage configuration steps
 func (st *Stage) configMainStage() {
+	sc := st.Scene
 	if st.NewWindow {
 		st.FullWindow = true
 	}
+	if st.NewWindow && st.Type == DialogStage && TheApp.Platform() == system.Web && st.Context != nil {
+		if st.Context.AsWidget().SizeClass() == SizeExpanded {
```
kkoreilly: Yes, you can make this `TheApp.Platform().IsMobile()` so that it works on native tablets/iPads etc. Another question is whether we should just require `!= SizeCompact` (to also include `SizeMedium`). Finally, this should be one if statement.

core/mainstage.go
```diff
@@ -398,5 +406,13 @@ func (sm *stages) mainHandleEvent(e events.Event) {
...
 			break
 		}
+		if st.Type == DialogStage { // modal dialog, by definition
+			if e.HasPos() {
+				b := st.Scene.SceneGeom.Bounds()
+				if e.WindowPos().In(b) { // don't propagate
+					break
+				}
+			}
+		}
```
kkoreilly, line 415: This is what I am talking about. Shouldn't this logic prevent the events from escaping from the modeless dialog? By the way, your comment is wrong, this is `modeless dialog, by definition`. And is `WindowPos` correct here (vs `Pos`)?

kkoreilly (changes requested):

core/mainstage.go
rcoreilly: done

core/events.go
```diff
@@ -398,18 +400,20 @@ func (em *Events) handlePosEvent(e events.Event) {
 			}
 			em.dragHovers = em.updateHovers(hovs, em.dragHovers, e, events.DragEnter, events.DragLeave)
 			em.dragMove(e)                              // updates sprite position
-			em.drag.AsWidget().HandleEvent(e)           // raw drag
```
kkoreilly, line 401: Why no longer handling raw drag?

core/mainstage.go
```diff
@@ -165,17 +165,23 @@ func (st *Stage) firstWindowStages() *stages {
...
+	sc := st.Scene
 	if st.NewWindow {
 		st.FullWindow = true
 	}
+	if st.NewWindow && st.Type == DialogStage && TheApp.Platform().IsMobile() && st.Context != nil && st.Context.AsWidget().SizeClass() != SizeCompact {
+		st.NewWindow = false
+		st.FullWindow = false
+		st.Modal = false
+		st.Scrim = false
+	}
```
kkoreilly: You can fold this if statement into the end of the if statement below to get rid of the duplicate mobile condition and `NewWindow` setting.

core/mainstage.go
```diff
@@ -398,5 +404,13 @@ func (sm *stages) mainHandleEvent(e events.Event) {
 		if e.IsHandled() || st.Modal || st.Type == WindowStage || st.FullWindow {
 			break
 		}
+		if st.Type == DialogStage { // modal dialog, by definition
```
kkoreilly: Again this is modeless, not modal.

core/style.go
```diff
@@ -225,4 +225,7 @@ func styleFromTags(w Widget, tags reflect.StructTag) {
 		setFromTag(tags, "grow", func(v float32) { s.Grow.X = v })
 		setFromTag(tags, "grow-y", func(v float32) { s.Grow.Y = v })
 	})
+	if v, ok := tags.Lookup("new-window"); ok && v == "+" {
```
kkoreilly: Just `if tags.Get("new-window") == "+"`. If it is `+`, it is also `ok`.

kkoreilly (changes requested):

core/events.go
rcoreilly: oops that was an accidental delete.

core/mainstage.go
rcoreilly: easier if you just do the logic the way you want it.

kkoreilly (approved):

PR #1167, opened 2024-08-21
document that you should re-install the core tool each time you update to a new version of Cogent Core

kkoreilly:


rcoreilly (approved):

rcoreilly:
Imagine how convenient it would be if you could just type `core selfupdate` instead of `go install cogentcore.org/core/cmd/core@main`

kkoreilly:
Once things are more stable, we will recommend that most people install `@latest`, but we will still install locally or `@main`, so again, it does not make sense to have a centralized command.

theclapp:
> Imagine how convenient it would be if you could just type `core selfupdate`

Failing that, it'd be nice if `core` could tell you that you needed to reinstall it.

Maybe you could publish a sample "post-merge" Git hook that would automatically "cd cmd/core ; go install" (or whatever) after a pull.

Or add "core pull" as a subcommand that did the same thing -- did a `git pull` and then reinstalled itself.

kkoreilly:
The issue is that most people shouldn't have to clone the `core` repo locally, so something based on `git pull` wouldn't work well. However, it might be feasible to make `core` commands detect when the version of the core tool differs from the latest and reinstall it for you (I'm not completely sure how we would detect the current version though). I think we can consider this more deeply once things are more stable and there is a clearer update path (`@latest`). This might also end up being part of #992.

PR #1176, opened 2024-08-24
properly deal with scrolling for RenderDraw in xyz, video

rcoreilly:
xyz was not scrolling properly due to logic issue in direct render draw, now fixed.

### DIFF
--- a/core/layout.go
+++ b/core/layout.go
@@ -1853,7 +1853,8 @@ func (fr *Frame) scrollResetIfNone() {
 // DirectRenderDrawBBoxes returns the destination and source bounding boxes
 // for RenderDraw call for widgets that do direct rendering.
 // The destBBox.Min point can be passed as the dp destination point for Draw
-// function, and srcBBox is
+// function, and srcBBox is the source region.  Empty flag indicates if either
+// of the srcBBox dimensions are <= 0.
 func (wb *WidgetBase) DirectRenderDrawBBoxes(srcFullBBox image.Rectangle) (destBBox, srcBBox image.Rectangle, empty bool) {
 	tbb := wb.Geom.TotalBBox
 	destBBox = tbb.Add(wb.Scene.SceneGeom.Pos)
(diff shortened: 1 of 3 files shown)


video/video.go
```diff
@@ -84,8 +84,21 @@ func (v *Video) RenderDraw(drw system.Drawer, op draw.Op) {
...
+		ibb.Min.X = tbb.Min.X - pos.X
+	}
+	if pos.Y < tbb.Min.Y {
+		ibb.Min.X = tbb.Min.Y - pos.X
+	}
+	sz := ibb.Size()
+	if sz.X <= 0 || sz.Y <= 0 {
+		return
+	}
+	drw.Scale(bb, frame, ibb, v.Rotation, draw.Src, unchanged)
```
kkoreilly: Is there a way to avoid forcing each widget to do all of this work manually?

rcoreilly: making a helper function..  could do that.

core/layout.go
```diff
@@ -1847,3 +1849,25 @@ func (fr *Frame) scrollResetIfNone() {
 		}
 	}
 }
+
+// DirectRenderDrawBBoxes returns the destination and source bounding boxes
+// for RenderDraw call for widgets that do direct rendering.
+// The destBBox.Min point can be passed as the dp destination point for Draw
+// function, and srcBBox is
+func (wb *WidgetBase) DirectRenderDrawBBoxes(srcFullBBox image.Rectangle) (destBBox, srcBBox image.Rectangle, empty bool) {
```
kkoreilly, line 1858: Your documentation trails off here. Also I thought we weren't using `DirectRender` terminology anymore?

kkoreilly (changes requested):

core/layout.go
rcoreilly: oops. DirectRender is still used for direct renderers -- but everyone just has a RenderDraw method.

kkoreilly (approved):

PR #1178, opened 2024-08-25
Tabs.NewTab returns the tab, to allow for customization

rcoreilly:


core/tabs.go
```diff
@@ -197,17 +197,17 @@ func (ts *Tabs) CurrentTab() (Widget, int) {
 // NewTab adds a new tab with the given label and returns the resulting tab frame.
```
kkoreilly: Update documentation to match new behavior.

core/tabs.go
```diff
@@ -197,17 +197,17 @@ func (ts *Tabs) CurrentTab() (Widget, int) {
...
 	fr := ts.getFrame()
 	idx := len(fr.Children)
-	frame := ts.insertNewTab(label, idx, icon...)
-	return frame
+	frame, button := ts.insertNewTab(label, idx, icon...)
+	return frame, button
 }
 
 // insertNewTab inserts a new tab with the given label at the given index position
 // within the list of tabs and returns the resulting tab frame. An optional icon
```
kkoreilly: Update doc.

core/tabs.go
```diff
@@ -197,17 +197,17 @@ func (ts *Tabs) CurrentTab() (Widget, int) {
 // NewTab adds a new tab with the given label and returns the resulting tab frame.
 // It is the main end-user API for creating new tabs. An optional icon can also
 // be passed for the tab button.
-func (ts *Tabs) NewTab(label string, icon ...icons.Icon) *Frame {
+func (ts *Tabs) NewTab(label string, icon ...icons.Icon) (*Frame, *Tab) {
```
kkoreilly: Stop taking icon for all of these functions and let people set it manually? This var arg icon setup is inconsistent with our other APIs, although maybe we could keep it if it is really helpful?

core/tabs_test.go
```diff
@@ -23,19 +23,19 @@ func TestTabs(t *testing.T) {
 func TestTabsWidgets(t *testing.T) {
 	b := NewBody()
 	ts := NewTabs(b)
-	first := ts.NewTab("First")
+	first, _ := ts.NewTab("First")
```
kkoreilly, line 26: I must say that I still really don't like the look of this `, _ :=` for basically every single use case. I wish we could have the same builtin map access paradigm where you can add the second return value only if you want it, although obviously that is not really idiomatic.

kkoreilly (changes requested):

core/tabs.go
rcoreilly: The main cases where icons are set is in tabs_test and the docs.  outside of nav-style tabs, how common is it to have icons?  browsers for sure for the favicon, but that certainly doesn't need the one-liner. 

Getting rid of the arg would definitely make it more cumbersome to set the icon, and there is a kind of logic that says that the label and / or icon are the defining properties of the tab, so you should set those at creation.

Overall, I'm in favor of keeping.

kkoreilly: We decided that it should not take the icon directly, as it does not need to and the icon is a purely aesthetic property, whereas the label is a necessary identifier. It is much more consistent with our other APIs to call `SetIcon` separately.

kkoreilly (approved):

PR #1180, opened 2024-08-26
Elide file button text

### LINKED ISSUE #1177: Suggestion on FileButton length after selection
### Describe the feature

Currently, when a file is selected with a FileButton, the button displays the path in full. The longer the path, longer the button; and when it reaches the limit of a set frame or window the path is cut off at the end so that the user wouldn't be able to verify the correct file anyway.
It would provide more flexibility with space consideration to shrink that to something like:
mountpoint/.../filename.ext
/...filename.ext
filename.ext

For example: "C:/.../filename" or "home/.../filename" would help keep the button from expanding too far.  For the convenience of being able to double check the path, (which I assume is why the button currently displays the full path,) Perhaps changing the tooltip for the button to the full path could keep that convenience.

As an afterthought, perhaps adding an option for a character limit could also be beneficial from a front end design standpoint.

### Relevant code

_No response_

kkoreilly:
Elide `FileButton` path text and add `WidgetTooltip` function for `FileButton` and `TreeButton` with the full un-elided path.

Fixes #1177

### DIFF
--- a/core/filepicker.go
+++ b/core/filepicker.go
@@ -6,6 +6,7 @@ package core
 
 import (
 	"fmt"
+	"image"
 	"log"
 	"log/slog"
 	"os"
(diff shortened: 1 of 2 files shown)


core/filepicker.go
```diff
@@ -731,7 +733,7 @@ func (fb *FileButton) Init() {
 		if fb.Filename == "" {
 			fb.SetText("Select file")
 		} else {
-			fb.SetText(fb.Filename)
+			fb.SetText(elide.Middle(fb.Filename, 38))
```
rcoreilly, line 736: so just a magic number 38 without any global var to override?

kkoreilly: We do the exact same for tree path eliding, and the full path is visible in the tooltip. We can add an option if someone requests it, but there is no point in adding additional complexity prematurely. Regardless, a field, not a global variable, would be the solution if ever necessary.

rcoreilly (approved):

PR #1181, opened 2024-08-26
several important fixes for xyz arrow rendering and line selection

rcoreilly:


xyz/node.go
```diff
@@ -107,6 +107,11 @@ type NodeBase struct {
 	// 2D bounding box for region occupied within Scene Frame that we render onto.
 	// This is BBox intersected with Frame bounds.
 	SceneBBox image.Rectangle `edit:"-" copier:"-" json:"-" xml:"-" set:"-"`
+
+	// IsLinear indicates that this element contains a line-like shape,
+	// which engages a more selective event processing logic to determine
+	// if the node was selected based on a mouse click point.
+	IsLinear bool
```
kkoreilly: Is this strictly necessary and does it need to be exported?

rcoreilly: yes absolutely necessary -- applies across different node types (groups and solids).  not 100% necessary to export -- I guess I'll follow the conservative logic for now.

kkoreilly (approved):

PR #1187, opened 2024-08-28
tensorcore.TensorEditor -- needed for ccn sims

rcoreilly:


kkoreilly (changes requested):
This repetition between `core.Table`, `tensorcore.Table`, and `tensorcore.TensorEditor` is getting out of hand.

rcoreilly:
If you look at the code for each, you will see that they each have unique logic and capabilities, and that the code for each is relatively concentrated on these differences, rather than the shared elements, especially compared to previous iterations.

please merge asap b/c this is needed for the class sims.

kkoreilly (approved):

PR #1191, opened 2024-08-29
filetree FilterFunc

rcoreilly:
Adds a FilterFunc to filetree.Tree to filter which files / dirs are shown in the tree.  Yaegi crashes on it so it doesn't work in the docs (commented that out with a TODO, will file an issue), but tested in cogent code to filter out the annoying .DS_Store files! :)


filetree/tree.go
```diff
@@ -47,6 +48,10 @@ type Tree struct {
 	// type of node to create; defaults to [Node] but can use custom node types
 	FileNodeType *types.Type `display:"-" json:"-" xml:"-"`
 
+	// function, if set, determines whether to include the given node in the tree.
```
kkoreilly: Newly written documentation should be of the form `IdentifierName ...`

kkoreilly (changes requested):

filetree/tree.go
```diff
@@ -37,37 +37,37 @@ type Tree struct {
...
+	// DirsOnTop indicates whether all directories are placed at the top of the tree.
 	// Otherwise everything is mixed.
 	DirsOnTop bool
 
-	// type of node to create; defaults to [Node] but can use custom node types
+	// FileNodeType is the type of node to create; defaults to [Node] but can use custom node types
 	FileNodeType *types.Type `display:"-" json:"-" xml:"-"`
 
-	// function, if set, determines whether to include the given node in the tree.
+	// FilterFunc if set, determines whether to include the given node in the tree.
```
kkoreilly: Sorry for nitpicking but there should still be a comma after `FilterFunc`.

kkoreilly (changes requested):

kkoreilly (approved):

PR #1193, opened 2024-08-30
fix simatgrid sizing for labels and replace Col -> Column for all pca…

rcoreilly:
… / svd / simat methods that still used Col


rcoreilly:
note: had to export FitSizeMax method for simmatgrid

kkoreilly (approved):

PR #1202, opened 2024-09-02
support Wrap mode in toolbar -- don't put in overflow

rcoreilly:
this is probably more effective for netview in emer sims, vs. overflow.  lots of active controls in that toolbar.


kkoreilly (commented):
I think we should consider whether we want to make this the default. Also, we should probably add some docs for this.

rcoreilly:
ok we can discuss that later..  but could you merge this PR now so I can have a complete updated ccn sims for this coming week?

kkoreilly (approved):

PR #1211, opened 2024-09-08
Fix List slice reflect panic and NodeBase.WalkDown panic

kkoreilly:
Also clean up style docs and reflectx slice code.

### DIFF
--- a/tree/nodebase.go
+++ b/tree/nodebase.go
@@ -497,7 +497,9 @@ func (n *NodeBase) WalkDown(fun func(n Node) bool) {
 outer:
 	for {
 		cb := cur.AsTree()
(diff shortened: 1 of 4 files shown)


tree/nodebase.go
```diff
@@ -497,7 +497,7 @@ func (n *NodeBase) WalkDown(fun func(n Node) bool) {
 outer:
 	for {
 		cb := cur.AsTree()
-		if cb.This != nil && fun(cur) { // false return means stop
+		if fun(cur) && cb.This != nil { // false return means stop
```
rcoreilly: better to tripple check: `cb.This != nil && fun(cur)  && cb.This != nil`

rcoreilly (approved):

PR #1212, opened 2024-09-08
trying no-max on row height

rcoreilly:
this seems to work but need more thorough testing -- will clean up later if it all looks good.


kkoreilly (approved):
Just to be clear, the purpose of this is to fix layout issues when the row height is not constant.

PR #1213, opened 2024-09-08
tensor generic compute functions

rcoreilly:
Per design discussion in: https://github.com/cogentcore/cogent/discussions/324

The `tensor.Indexed` type provides the universal representation of a homogenous data type throughout all the packages here, from scalar to vector, matrix, and beyond, because it can efficiently represent any kind of element with sufficient flexibility to enable a huge range of computations to be elegantly expressed.  The indexes provide a specific view onto the underlying [Tensor] data, applying to the outermost _row_ dimension (with default row-major indexing).  For example, sorting and filtering a tensor only requires updating the indexes while doing nothing to the Tensor itself.

The `Vectorize` function and its variants provide a universal "apply function to tensor data" mechanism (often called a "map" function, but that name is already taken in Go).  It takes an `N` function that determines how many indexes to iterate over (and this function can also do any initialization prior to iterating), a compute function that gets an index and a list of tensors, which is applied to every index, and a varargs list of indexed tensors.  It is completely up to the compute function how to interpret the index.  There is a Threaded version of this for parallelizable functions, and a GPU version.

All tensor package functions are registered using a global name-to-function map (`Funcs`), and can be called by name via `tensor.Call` or `tensor.CallOut` (creates the appropriate output tensors for you). Standard enumerated functions in `stats` and `metrics` have a `FuncName` method that appends the package name, which is how they are registered and called.

A `Table` automatically supplies a shared list of row Indexes for its `Indexed` columns, efficiently allowing all the heterogeneous data columns to be sorted and filtered together.

Added: [tmath](tmath) implements all standard math functions on `tensor.Indexed` data, including the standard `+, -, *, /` operators.  `cosl` then calls these functions.

Major cleanup of: `stats`, which implements a number of different ways of analyzing tensor and table data, including:
    - [cluster](cluster) implements agglomerative clustering of items based on [metric](metric) distance / similarity matrix data.
    - [convolve](convolve) convolves data (e.g., for smoothing).
    - [glm](glm) fits a general linear model for one or more dependent variables as a function of one or more independent variables.  This encompasses all forms of regression.
    - [histogram](histogram) bins data into groups and reports the frequency of elements in the bins.
    - [metric](metric) computes similarity / distance metrics for comparing two tensors, and associated distance / similarity matrix functions, including PCA and SVD analysis functions that operate on a covariance matrix.
    - [stats](stats) provides a set of standard summary statistics on a range of different data types, including basic slices of floats, to tensor and table data.  It also includes the ability to extract Groups of values and generate statistics for each group, as in a "pivot table" in a spreadsheet.


base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
+// Copyright (c) 2024, Cogent Core. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+package keylist implements an ordered list (slice) of items,
```
kkoreilly: `Package` not `package`.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+through the tuple is the major problem with ordmap.
+*/
+package keylist
+
+import (
+	"fmt"
+	"slices"
+)
+
+// TODO: probably want to consolidate ordmap and keylist
```
kkoreilly: Link to #1224.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+
+// List implements an ordered list (slice) of Values,
+// with a map from a key (e.g., names) to indexes,
+// to support fast lookup by name.
+type List[K comparable, V any] struct { //types:add
+	// List is the ordered slice of items.
+	Values []V
+
+	Keys []K
```
kkoreilly, line 31: This needs to be exported? And add documentation.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// indexes is the key-to-index mapping.
+	indexes map[K]int
+}
+
+// New returns a new Key List.  Zero value
+// is usable without initialization, so this is
+// just a simple standard convenience method.
+func New[K comparable, V any]() *List[K, V] {
```
kkoreilly, line 40: `New returns a new [List]. The zero value...` Also, if this doesn't actually do anything and the zero value is usable, we should just get rid of it.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+	return &List[K, V]{}
+}
+
+func (kl *List[K, V]) newIndexes() {
```
kkoreilly: You should probably call this `makeIndexes` to be consistent with the `make` function that you are calling.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+// Reset resets the list, removing any existing elements.
+func (kl *List[K, V]) Reset() {
+	kl.Values = nil
+	kl.Keys = nil
+	kl.newIndexes()
+}
+
+// Add adds an item to the list with given key.
+// An error is returned if the key is already on the list.
+// See [AddReplace] for a method that automatically replaces.
```
kkoreilly: `List.AddReplace`, not `AddReplace`.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	kl.indexes[key] = len(kl.Values)
+	kl.Values = append(kl.Values, val)
+	kl.Keys = append(kl.Keys, key)
+	return nil
+}
+
+// AddReplace adds an item to the list with given key,
+// replacing any existing item with the same key.
+func (kl *List[K, V]) AddReplace(key K, val V) {
```
kkoreilly: Is it worth having these separate APIs and introducing errors? If it is really necessary, it should at least be `Add` and `AddTry`.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+	kl.indexes[key] = len(kl.Values)
+	kl.Values = append(kl.Values, val)
+	kl.Keys = append(kl.Keys, key)
+}
+
+// Insert inserts the given value with the given key at the given index.
+// This is relatively slow because it needs regenerate the keys list.
+// It returns an error if the key already exists because
+// the behavior is undefined in that situation.
+func (kl *List[K, V]) Insert(idx int, key K, val V) error {
```
kkoreilly: I really don't like all of these functions returning errors, which you probably won't handle anyway; panic is standard for unspecified behavior in low-level data types.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+	for i, k := range kl.Keys {
+		kl.indexes[k] = i
+	}
+	return nil
+}
+
+// ValueByKey returns the value corresponding to the given key,
+// with a zero value returned for a missing key. See [List.ValueByKeyTry]
+// for one that returns a bool for missing keys.
+func (kl *List[K, V]) ValueByKey(key K) V {
```
kkoreilly: Given that `ValueByKey` is one of the most common use cases, I feel like we should have a shorter name.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+	if kl == nil {
+		return 0
+	}
+	return len(kl.Values)
+}
+
+// DeleteIndex deletes item(s) within the index range [i:j].
+// This is relatively slow because it needs to regenerate the
+// index map.
+func (kl *List[K, V]) DeleteIndex(i, j int) {
```
kkoreilly: Just call this `Delete`?

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+func (kl *List[K, V]) DeleteKey(key K) bool {
+	idx, ok := kl.indexes[key]
+	if !ok {
+		return false
+	}
+	kl.DeleteIndex(idx, idx+1)
+	return true
+}
+
+// Copy copies all of the entries from the given keyed list
```
kkoreilly: I specifically raised concerns that `keylist` would be confused with `KeyedList`, and you rejected them yet are making precisely the same mistake.

base/keylist/keylist.go
```diff
@@ -0,0 +1,218 @@
...
+// into this list. It keeps existing entries in this
+// list unless they also exist in the given list, in which case
+// they are overwritten.  Use [Reset] first to get an exact copy.
```
kkoreilly: Again, you need to specify the method receiver in the doc comment.

base/metadata/metadata.go
```diff
@@ -52,3 +64,23 @@ func (md *Data) Copy(src Data) {
...
 	maps.Copy(*md, src)
 }
+
+// SetName sets the "Name" standard key.
+func (md *Data) SetName(name string) {
+	md.Set("Name", name)
+}
+
+// GetName returns the "Name" standard key value (empty if not set).
+func (md *Data) GetName() string {
```
kkoreilly: Typically you omit the `Get` in Go function names.

base/metadata/metadata.go
```diff
@@ -52,3 +64,23 @@ func (md *Data) Copy(src Data) {
...
+	return errors.Ignore1(Get[string](*md, "Name"))
+}
+
+// SetDoc sets the "Doc" standard key.
+func (md *Data) SetDoc(doc string) {
+	md.Set("Doc", doc)
+}
+
+// GetDoc returns the "Doc" standard key value (empty if not set).
+func (md *Data) GetDoc() string {
```
kkoreilly: Again, omit the `Get`.

base/ordmap/ordmap.go
```diff
@@ -35,7 +35,6 @@ type KeyValue[K comparable, V any] struct {
 // and the fast key lookup of a map. A map stores an index
 // into a slice that has the value and key associated with the value.
 type Map[K comparable, V any] struct {
-
```
kkoreilly: We do like to have this newline.

filetree/node.go
```diff
@@ -337,7 +337,7 @@ func (fn *Node) dirFileList() []fs.FileInfo {
 
 func sortByModTime(files []fs.FileInfo) {
 	slices.SortFunc(files, func(a, b fs.FileInfo) int {
-		return a.ModTime().Compare(b.ModTime())
+		return -a.ModTime().Compare(b.ModTime())
```
kkoreilly: Just put b before a instead of using negative?

math32/math.go
```diff
@@ -804,6 +804,17 @@ func ClampInt(x, a, b int) int {
 	return x
 }
 
+// Clamp64 clamps x to the provided closed interval [a, b]
+func Clamp64(x, a, b float64) float64 {
```
kkoreilly: Please just be reasonable and make **one** `math32.Clamp` generic function. All scalar `math32` functions should be generic (moreover, I really think the whole package should be generic and not called `math32`).

plot/plotcore/barplot.go
```diff
@@ -35,16 +32,16 @@ func (pl *PlotEditor) genPlotBar() {
...
+		lcol := pl.table.ColumnIndex(pl.Options.Legend)
+		if lcol < 0 {
+			log.Println("plot.Legend not found: " + pl.Options.Legend)
 		} else {
-			xview.SortColumnNames([]string{pl.Options.Legend, xp.Column}, table.Ascending) // make it fit!
-			lsplit = split.GroupBy(xview, pl.Options.Legend)
-			nleg = max(lsplit.Len(), 1)
+			// xview.SortColumnNames([]string{pl.Options.Legend, xp.Column}, tensor.Ascending) // make it fit!
+			// lsplit = split.GroupBy(xview, pl.Options.Legend)
+			// nleg = max(lsplit.Len(), 1)
```
kkoreilly, line 44: Why are there so many things commented out in this function? What is your plan? Can you just get rid of them?

plot/plotcore/metadata.go
```diff
@@ -0,0 +1,122 @@
...
+// also the Column name and IsString should not be in the struct!
+
+// fromMeta sets plot options from meta data.
+func (po *PlotOptions) fromMeta(dt *table.Table) {
+	po.FromMetaMap(dt.Meta)
+}
+
+// FromMetaMap sets plot options from meta data map.
+func (po *PlotOptions) FromMetaMap(meta metadata.Data) {
+	if typ, err := metadata.Get[string](meta, "Type"); err == nil {
```
kkoreilly, line 26: You could write a nice DRY helper function for all of this stuff...

plot/plotcore/metadata.go
```diff
@@ -0,0 +1,122 @@
...
+		po.XAxisLabel = lb
+	}
+	if lb, err := metadata.Get[string](meta, "YAxisLabel"); err == nil {
+		po.YAxisLabel = lb
+	}
+}
+
+// fromMetaMap sets column options from meta data map.
+func (co *ColumnOptions) fromMetaMap(meta metadata.Data) {
+	if op, err := metadata.Get[bool](meta, co.Column+":On"); err == nil {
```
kkoreilly, line 69: We are using colons as separators?

plot/plotcore/metadata.go
```diff
@@ -0,0 +1,122 @@
...
+func PlotColumnZeroOne() (*ColumnOptions, string) {
+	opts := &ColumnOptions{}
+	opts.Range.SetMin(0)
+	opts.Range.SetMax(1)
+	return opts, "PlotColumnOptions"
+}
+
+// SetPlotColumnOptions sets given plotting options for named items
+// within this directory, in Metadata.
+func SetPlotColumnOptions(md metadata.Data, opts *ColumnOptions) {
```
kkoreilly, line 115: Why does this need to be a global function? (same with `PlotColumnOptions` below)

plot/plotcore/xyplot.go
```diff
@@ -28,16 +25,16 @@ func (pl *PlotEditor) genPlotXY() {
...
+		ix := pl.table.ColumnIndex(pl.Options.Legend)
+		if ix < 0 {
 			slog.Error("plot.Legend", "err", err.Error())
 		} else {
-			errors.Log(xview.SortStableColumnNames([]string{pl.Options.Legend, xp.Column}, table.Ascending))
-			lsplit = split.GroupBy(xview, pl.Options.Legend)
-			nleg = max(lsplit.Len(), 1)
+			// errors.Log(xview.SortStableColumnNames([]string{pl.Options.Legend, xp.Column}, table.Ascending))
+			// lsplit = split.GroupBy(xview, pl.Options.Legend)
+			// nleg = max(lsplit.Len(), 1)
```
kkoreilly, line 37: Same question as `barplot.go` about commented out code.

kkoreilly (changes requested):
Some initial comments up to but not including `tensor`.

base/keylist/keylist.go
rcoreilly: `At`, `Value`, `ByKey` ?  I prefer `At`

base/keylist/keylist.go
rcoreilly: `DeleteByIndex`

rcoreilly: and `DeleteByKey`

plot/plotcore/barplot.go
rcoreilly: i will come back to that code later; uses new stats.Group code into datafs and it requires just a bit more time than I have now.. :)

plot/plotcore/metadata.go
rcoreilly: this entire heap of garbage will be replaced with an automatic field setting thing.

plot/plotcore/xyplot.go
rcoreilly: same answer.

plot/plotcore/metadata.go
rcoreilly: this is to be used by random other objects (e.g., tensors, tables) and is not part of the ColumnOptions api per se.

plot/plotcore/metadata.go
rcoreilly: next time.

tensor/base.go
```diff
@@ -148,47 +188,63 @@ func (tsr *Base[T]) SymmetricDim() int {
...
-func (tsr *Base[T]) MetaData(key string) (string, bool) {
-	if tsr.Meta == nil {
-		return "", false
-	}
-	val, ok := tsr.Meta[key]
-	return val, ok
+// Sprint returns a string representation of the given tensor,
+// with a maximum length of as given: output is terminated
+// when it exceeds that length. If maxLen = 0, [MaxSprintLength] is used.
+func Sprint(tsr Tensor, maxLen int) string {
```
kkoreilly, line 201: Is `Sprint` the best name for this?

tensor/bitslice/bitslice.go
```diff
@@ -76,7 +76,7 @@ func (bs *Slice) SetLen(ln int) {
 }
 
 // Set sets value of given bit index -- no extra range checking is performed -- will panic if out of range
-func (bs *Slice) Set(idx int, val bool) {
+func (bs *Slice) Set(val bool, idx int) {
```
kkoreilly, line 79: I'm not particularly happy about backwards setting; I also still don't like this for bit flags... I guess there isn't much choice given the need for var args in some contexts?

tensor/databrowser/filetree.go
```diff
@@ -41,7 +40,7 @@ func (fn *FileNode) WidgetTooltip(pos image.Point) (string, image.Point) {
 		switch fn.Info.Known {
 		case fileinfo.Number, fileinfo.String:
 			dv := DataFS(ofn)
-			v, _ := dv.AsString()
+			v := dv.AsString()
```
kkoreilly, line 43: What is `AsString` here? Why isn't it `String`?

tensor/cmd/tablecat/tablecat.go
```diff
@@ -110,13 +110,13 @@ func RawCat(files []string) {
 func AvgCat(files []string) {
 	dts := make([]*table.Table, 0, len(files))
 	for _, fn := range files {
-		dt := &table.Table{}
-		err := dt.OpenCSV(core.Filename(fn), table.Tab)
+		dt := table.NewTable()
```
kkoreilly, line 113: What about calling this `table.New()`?

tensor/datafs/data.go
```diff
@@ -28,20 +29,26 @@ type Data struct {
...
-	// Value is the underlying value of data;
-	// is a map[string]*Data for directories.
-	Value any
+	// Dir is for directory nodes, with all the items in the directory.
+	Dir *Dir
+
+	// DirTable is a summary [table.Table] with columns comprised of
+	// Value items in the directory, which can be used for plotting or other
+	// operations.
+	DirTable *table.Table
```
kkoreilly, line 44: This needs to be exported and stored? (as opposed to fetched by some function)

tensor/datafs/data.go
```diff
@@ -50,264 +57,135 @@ func NewData(dir *Data, name string) (*Data, error) {
...
-		d.Value = float32(v)
-		return true
-	}
-	if tsr := d.AsTensor(); tsr != nil {
-		return false
-	}
-	if dt := d.AsTable(); dt != nil {
-		return false
+// AsString returns data as scalar string.
+func (d *Data) AsString() string {
```
kkoreilly, line 136: Again, why isn't this called `String`?

tensor/datafs/data.go
```diff
@@ -50,264 +57,135 @@ func NewData(dir *Data, name string) (*Data, error) {
...
+	}
+	d.Data.SetStringRow(v, 0)
 }
 
-// SetFloat32 sets data from given float32 if it is a scalar value
-// that can be so set.  Returns false if not.
-func (d *Data) SetFloat32(v float32) bool {
-	return d.SetFloat64(float64(v))
+// AsFloat64 returns data as a scalar float64 (first element of tensor).
+func (d *Data) AsFloat64() float64 {
```
kkoreilly, line 152: Similarly why isn't this called `Float64`?

base/keylist/keylist.go
rcoreilly: yes needs to be exported; docs added.

base/keylist/keylist.go
rcoreilly: better to always have New functions to have a consistent interface for everything.  some need init, some don't but the api user shouldn't need to worry about it, and this insulates against any future possible change.

base/keylist/keylist.go
rcoreilly: `Set` is the new name for the auto-replace version, like a Go map.  `Add` always adds, so if it is already on the list, it is an error (current implemented behavior).

base/keylist/keylist.go
rcoreilly: ok, turned into a panic.

base/keylist/keylist.go
rcoreilly: `At` done.

plot/plotcore/barplot.go
rcoreilly: will do later.

plot/plotcore/metadata.go
rcoreilly: will do later.

tensor/datafs/data.go
kkoreilly: Same with `AsFloat32` and `AsInt` below.

tensor/datafs/list.go
```diff
@@ -0,0 +1,60 @@
...
+package datafs
+
+import "strings"
+
+const (
+	Short = false
+	Long  = true
+
+	DirOnly   = false
+	Recursive = true
```
kkoreilly, line 14: These bool constants are undocumented and ambiguously named. Moreover, the general bool constant paradigm is unideal, and given that you need to make this a cli config struct at some point anyway, I would strongly reconsider this approach.

tensor/datafs/list.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// todo: list options string
+
+func (d *Data) String() string {
```
kkoreilly, line 19: Is this `String` or `Label`?

tensor/datafs/list.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (d *Data) List(long, recursive bool) string {
+	if long {
+		return d.ListLong(recursive, 0)
+	}
+	return d.ListShort(recursive, 0)
+}
+
+func (d *Data) ListShort(recursive bool, indent int) string {
```
kkoreilly, line 33: If you are going to have these ridiculous bool constants, these shouldn't be exported.

tensor/examples/datafs-sim/sim.go
```diff
@@ -26,9 +26,9 @@ type Sim struct {
 
 // ConfigAll configures the sim
 func (ss *Sim) ConfigAll() {
-	ss.Root = errors.Log1(datafs.NewDir("Root"))
-	ss.Config = errors.Log1(ss.Root.Mkdir("Config"))
-	errors.Log1(datafs.New[int](ss.Config, "NRun", "NEpoch", "NTrial"))
+	ss.Root, _ = datafs.NewDir("Root")
+	ss.Config, _ = ss.Root.Mkdir("Config")
```
kkoreilly, line 30: Why are you ignoring these errors?

tensor/examples/planets/planets.go
```diff
@@ -0,0 +1,102 @@
...
+	"cogentcore.org/core/tensor/stats/stats"
+	"cogentcore.org/core/tensor/table"
+	"cogentcore.org/core/tree"
+)
+
+//go:embed *.csv
+var csv embed.FS
+
+// AnalyzePlanets analyzes planets.csv data following some of the examples
+// given here, using pandas:
```
kkoreilly, line 26: In what sense is this `using pandas`?

tensor/funcs.go
```diff
@@ -0,0 +1,286 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+
+	"cogentcore.org/core/base/errors"
+	"cogentcore.org/core/base/metadata"
+)
+
+// StringFirstArg should be used to set StringFirst functions
+const StringFirstArg = true
```
kkoreilly, line 17: Again, these bool constants are really not great, nor does it seem like this is a particularly well structured system if you have things like this.

tensor/funcs.go
```diff
@@ -0,0 +1,286 @@
...
+// in specific packages to call functions by enum String() names.
+// Use the standard Go CamelCase name -- will be auto-lowercased.
+// The number of output arguments must be provided here,
+// along with an optional first string argument if present;
+// the number of input arguments is automatically set from that.
+func AddFunc(name string, fun any, out int, stringFirst ...bool) error {
+	if Funcs == nil {
+		Funcs = make(map[string]*Func)
+	}
+	nm := strings.ToLower(name)
```
kkoreilly, line 79: Do `strcase.ToLowerCamel`, right?

kkoreilly: And update the documentation above accordingly.

tensor/funcs.go
```diff
@@ -0,0 +1,286 @@
...
+	fn, ok := Funcs[nm]
+	if !ok {
+		return nil, errors.Log(fmt.Errorf("tensor.CallOut: function of name %q not registered", name))
+	}
+	return fn.CallOut(tsr...)
+}
+
+// ArgCount returns the number of tensor arguments the function takes,
+// using a type switch.
+func (fn *Func) ArgCount() int {
```
kkoreilly, line 139: Why not just use reflection?

tensor/funcs.go
```diff
@@ -0,0 +1,286 @@
...
+		nargs = 4
+	case func(s string, a, b, c, d, e *Indexed):
+		nargs = 5
+	}
+	return nargs
+}
+
+// ArgCheck returns an error if the number of args in list does not
+// match the number required as specified.
+func (fn *Func) ArgCheck(n int, tsr ...*Indexed) error {
```
kkoreilly, line 169: This needs to be exported?

tensor/indexed.go
```diff
@@ -0,0 +1,706 @@
...
+}
+
+const (
+	// Ascending specifies an ascending sort direction for tensor Sort routines
+	Ascending = true
+
+	// Descending specifies a descending sort direction for tensor Sort routines
+	Descending = false
+
+	//	Stable specifies using stable, original order-preserving sort, which is slower.
```
kkoreilly, line 221: Your tab formatting is bad here. Also, these bool constants are once again pretty ambiguously named...

tensor/indexed.go
```diff
@@ -0,0 +1,706 @@
...
+// numerical order, producing the native ordering, while preserving
+// any filtering that might have occurred.
+func (ix *Indexed) SortIndexes() {
+	if ix.Indexes == nil {
+		return
+	}
+	sort.Ints(ix.Indexes)
+}
+
+func CompareAscending[T cmp.Ordered](a, b T, ascending bool) int {
```
kkoreilly, line 250: Unexport and/or document?

tensor/indexed.go
kkoreilly: And it doesn't even look like you use `Stable` and `Unstable`?

tensor/indexed.go
```diff
@@ -0,0 +1,706 @@
...
+		}
+	}
+}
+
+// Named arg values for FilterString
+const (
+	// Include means include matches
+	Include = false
+	// Exclude means exclude matches
+	Exclude = true
```
kkoreilly, line 322: Again, ambiguous names for bool constants.

tensor/indexed.go
```diff
@@ -0,0 +1,706 @@
...
+	UseCase = false
+)
+
+// FilterString filters the indexes using string values compared to given
+// string. Includes rows with matching values unless exclude is set.
+// If contains, only checks if row contains string; if ignoreCase, ignores case.
+// Use the named const args [Include], [Exclude], [Contains], [Equals],
+// [IgnoreCase], [UseCase] for greater clarity.
+// Uses first cell of higher dimensional data.
+func (ix *Indexed) FilterString(str string, exclude, contains, ignoreCase bool) { //types:add
```
kkoreilly, line 339: I would really prefer some sort of struct... Once you have a bunch of arguments with named bool constants, you should really just be using a struct.

kkoreilly (changes requested):
More comments through and including `tensor/indexed.go`.

tensor/io.go
```diff
@@ -11,14 +11,57 @@ import (
...
+// SetPrecision sets the "precision" metadata value that determines
+// the precision to use in writing floating point numbers to files.
+func SetPrecision(md metadata.Data, prec int) {
+	md.Set("precision", prec)
+}
+
+// GetPrecision gets the "precision" metadata value that determines
+// the precision to use in writing floating point numbers to files.
+// returns an error if not set.
+func GetPrecision(md metadata.Data) (int, error) {
```
kkoreilly, line 56: No `Get`.

tensor/number.go
```diff
@@ -36,41 +34,69 @@ type Int32 = Number[int32]
 // Byte is an alias for Number[byte].
 type Byte = Number[byte]
 
-// NewFloat32 returns a new Float32 tensor
+// NewFloat32 returns a new [Float32] tensor
 // with the given sizes per dimension (shape), and optional dimension names.
-func NewFloat32(sizes []int, names ...string) *Float32 {
-	return New[float32](sizes, names...).(*Float32)
+func NewFloat32(sizes ...int) *Float32 {
```
kkoreilly, line 39: Update doc; no more names.

kkoreilly: Also applies to many other functions below; do a find and replace to get rid of it.

tensor/shape.go
```diff
@@ -15,32 +17,37 @@ import (
 // Per C / Go / Python conventions, indexes are Row-Major, ordered from
 // outer to inner left-to-right, so the inner-most is right-most.
 type Shape struct {
-
```
kkoreilly, line 18: Keep the newlines!

tensor/shape.go
```diff
@@ -197,7 +232,7 @@ func RowMajorStrides(sizes []int) []int {
 
 // ColMajorStrides returns strides for sizes where the first dimension is inner-most
 // and subsequent dimensions are progressively outer
-func ColMajorStrides(sizes []int) []int {
+func ColMajorStrides(sizes ...int) []int {
```
kkoreilly, line 235: Rename `Col` to `Column`?

tensor/slice.go
```diff
@@ -0,0 +1,186 @@
...
+	"slices"
+)
+
+// Range represents a range of values, for extracting slices of data,
+// using standard for loop logic with a Start and exclusive End value,
+// and an increment: for i := Start; i < End; i += Incr.
+// The zero value means all values in the dimension.
+type Range struct {
+	// Starting value.
+	Start int
```
kkoreilly, line 18: Start doc comments with the identifier name.

tensor/slice.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// End value. 0 default = size of relevant dimension.
+	End int
+
+	// Increment. must be positive, 1 or greater. 0 default = 1.
+	Incr int
```
kkoreilly, line 24: Spell out `Increment`?

kkoreilly: Or use `Step` or something similar?

tensor/slice.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// EndActual is the actual end value given the size of the dimension.
+func (rn *Range) EndActual(size int) int {
```
kkoreilly, line 28: `GetEnd` instead of `EndActual` would be more typical (same for `IncrActual` below).

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+// to a dir, then cal it.
+
+// Node is one node in the cluster
+type Node struct {
+	// index into original distance matrix; only valid for for terminal leaves.
+	Index int
+
+	// Distance value for this node, i.e., how far apart were all the kids from
+	// each other when this node was created. is 0 for leaf nodes
+	Dist float64
```
kkoreilly, line 31: `Distance`?

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// ParDist is total aggregate distance from parents; The X axis offset at which our cluster starts.
+	ParDist float64
```
kkoreilly, line 34: `ParentDistance`?

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"cogentcore.org/core/base/indent"
+	"cogentcore.org/core/tensor"
+	"cogentcore.org/core/tensor/stats/stats"
+)
+
+// todo: all of this data goes into the datafs
+// Cluster makes a new dir, stuffs results in there!
+// need a global "cwd" that it uses, so basically you cd
+// to a dir, then cal it.
```
kkoreilly, line 22: `call`?

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Node is one node in the cluster
+type Node struct {
```
kkoreilly, line 25: Why isn't this a `tree.Node`?

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// ParDist is total aggregate distance from parents; The X axis offset at which our cluster starts.
+	ParDist float64
+
+	// Y is y-axis value for this node; if a parent, it is the average of its kids Y's,
+	// otherwise it counts down.
+	Y float64
+
+	// Kids are child nodes under this one.
+	Kids []*Node
```
kkoreilly, line 41: `Children`?

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// IsLeaf returns true if node is a leaf of the tree with no kids
+func (nn *Node) IsLeaf() bool {
```
kkoreilly, line 45: `HasChildren`?

tensor/stats/cluster/cluster.go
```diff
@@ -0,0 +1,163 @@
...
+	return len(nn.Kids) == 0
+}
+
+// Sprint prints to string
+func (nn *Node) Sprint(labels *tensor.Indexed, depth int) string {
```
kkoreilly, line 50: `String`?

tensor/stats/metric/funcs.go
```diff
@@ -0,0 +1,329 @@
...
+// and use the min of the length of the two tensors.
+// Metric functions cannot be computed in parallel,
+// e.g., using VectorizeThreaded or GPU, due to shared writing
+// to the same output values.  Special implementations are required
+// if that is needed.
+type MetricFunc func(a, b, out *tensor.Indexed)
+
+// SumSquaresScaleFuncOut64 computes the sum of squares differences between tensor values,
+// returning scale and ss factors aggregated separately for better numerical stability, per BLAS.
+func SumSquaresScaleFuncOut64(a, b, out *tensor.Indexed) (scale64, ss64 *tensor.Indexed) {
```
kkoreilly, line 35: Every one of these needs a `Func` suffix?

tensor/stats/metric/vec.go
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) 2024, Cogent Core. All rights reserved.
```
kkoreilly, line 1: Call this file `vector.go`.

tensor/stats/metric/vec.go
```diff
@@ -0,0 +1,202 @@
...
+	out.Tensor.SetShape(osh.Sizes...)
+	out.Tensor.SetNames(osh.Names...)
+	out.Indexes = nil
+	return min(na, nb)
+}
+
+// VecFunc is a helper function for metrics functions, dealing with iterating over
+// the Cell subspace per row and initializing the aggregation values for first index.
+// It also skips over NaN missing values.
+func VecFunc(idx int, a, b, out *tensor.Indexed, ini float64, fun func(a, b, agg float64) float64) {
```
kkoreilly, line 65: `VectorFunc`?

tensor/stats/metric/vec.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		out.SetFloat1D(fun(av, bv, out.Float1D(i)), i)
+	}
+}
+
+// VecSSFunc is a helper function for metric functions, dealing with iterating over
+// the Cell subspace per row and initializing the aggregation values for first index.
+// This version does sum-of-squares integration over 2 output vectors,
+// It also skips over NaN missing values.
+func VecSSFunc(idx int, a, b, out1, out2 *tensor.Indexed, ini1, ini2 float64, fun func(a, b float64) float64) {
```
kkoreilly, line 87: `VectorSSFunc`?

kkoreilly: `Vector` instead of `Vec` for the rest of the functions in this file?

tensor/stats/stats/funcs.go
```diff
@@ -4,62 +4,341 @@
...
+			out.SetFloat1D(math.Sqrt(var64.Float1D(i)), i)
+		} else {
+			out.SetFloat1D(math.Sqrt(var64.Float1D(i))/math.Sqrt(c), i)
+		}
+	}
+}
+
+// VarPopFuncOut64 computes the population variance of tensor values.
+// and returns the Float64 output values for subsequent use.
+func VarPopFuncOut64(in, out *tensor.Indexed) (var64, mean64, count64 *tensor.Indexed) {
```
kkoreilly, line 222: There is a question as to whether you would be better off spelling out these names some more...

tensor/stats/stats/funcs.go
```diff
@@ -4,62 +4,341 @@
...
+// so each cell in a higher-dimensional output reflects the _row-wise_
+// stat for that cell across the different rows.  To compute a stat
+// on the [tensor.SubSpace] cells themselves, must call on a
+// [tensor.New1DViewOf] the sub space.
+// All stats functions skip over NaN's, as a missing value.
+// Stats functions cannot be computed in parallel,
+// e.g., using VectorizeThreaded or GPU, due to shared writing
+// to the same output values.  Special implementations are required
+// if that is needed.
+type StatsFunc func(in, out *tensor.Indexed)
```
kkoreilly, line 28: We are in `stats`; what about just `Func`?

tensor/stats/stats/funcs.go
```diff
@@ -4,62 +4,341 @@
...
+	}, in, out)
 }
 
-// MaxAbsFunc is an StatFunc that computes a max aggregate.
-// use -math.MaxFloat64 for initial agg value.
-func MaxAbsFunc(idx int, val float64, agg float64) float64 {
-	return math.Max(agg, math.Abs(val))
+// MinFunc computes the min of tensor values.
+// See [StatsFunc] for general information.
+func MinFunc(in, out *tensor.Indexed) {
```
kkoreilly, line 85: Still not a fan of all of the `Func` suffixes.

kkoreilly (changes requested):
More comments through and including `tensor/stats/stats/quantiles.go`

tensor/stats/stats/README.md
```diff
@@ -1,36 +1,92 @@
 # stats
```
kkoreilly, line 1: Does this really need to be double nested? (`stats/stats`)

tensor/stats/stats/stats.go
```diff
@@ -4,70 +4,133 @@
...
+}
+
 // Stats is a list of different standard aggregation functions, which can be used
 // to choose an aggregation function
 type Stats int32 //enums:enum
 
 const (
-	// count of number of elements
+	// count of number of elements.
 	Count Stats = iota
```
kkoreilly, line 46: If you're going to add the trailing period, which is good, you should also add the identifier name at the start.

tensor/stats/stats/stats.go
```diff
@@ -4,70 +4,133 @@
...
+	spl := strings.Split(name, ".")
+	if len(spl) > 1 {
+		return spl[len(spl)-1]
+	}
+	return name
+}
+
+// Stat calls a standard Stats enum function on given tensors.
+// Output results are in the out tensor.
+func Stat(stat Stats, in, out *tensor.Indexed) {
```
kkoreilly, line 128: Again, we are in `stats`. What about `stats.Call` as a clearer name?

tensor/stats/stats/vec.go
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2024, Cogent Core. All rights reserved.
```
kkoreilly, line 1: Again, call this file `vector.go`.

tensor/stats/stats/vec.go
```diff
@@ -0,0 +1,147 @@
...
+	out.Tensor.SetShape(osh.Sizes...)
+	out.Tensor.SetNames(osh.Names...)
+	out.Indexes = nil
+	return n
+}
+
+// VecFunc is a helper function for stats functions, dealing with iterating over
+// the Cell subspace per row and initializing the aggregation values for first index.
+// It also skips over NaN missing values.
+func VecFunc(idx int, in, out *tensor.Indexed, ini float64, fun func(val, agg float64) float64) {
```
kkoreilly, line 93: `Vector`, not `Vec`.

tensor/stats/stats/vec.go
```diff
@@ -0,0 +1,147 @@
...
+		out.SetFloat1D(fun(val, out.Float1D(i)), i)
+	}
+}
+
+// Vec2inFunc is a helper function for stats functions, dealing with iterating over
+// the Cell subspace per row and initializing the aggregation values for first index.
+// This version has 2 input vectors, the second input being the output of another stat
+// e.g., the mean for Var.
+// It also skips over NaN missing values.
+func Vec2inFunc(idx int, in1, in2, out *tensor.Indexed, ini float64, fun func(val1, val2, agg float64) float64) {
```
kkoreilly, line 112: `In` should be capitalized in this name (same with `Out` below, and probably true in other cases; do a global find and replace).

kkoreilly (changes requested):
More comments through and including `tensor/string.go`.

tensor/bitslice/bitslice.go
rcoreilly: yeah it is essential for indexes..

tensor/datafs/data.go
rcoreilly: yeah it is persistent -- there is some cost to making it, and it has pointers to data columns.

tensor/examples/datafs-sim/sim.go
rcoreilly: this is just an end-user test., and doesn't have any errors..

tensor/base.go
rcoreilly: yes, unless you have a better suggestion?

kkoreilly: `String`? `AsString`? `ToString`?

tensor/databrowser/filetree.go
rcoreilly: `String` as you know is the `fmt.Stringer` function, but this is _not_ that -- is is accessing the `Data` value representation as a string.  The two have very different usages.  For this reason, it probably also makes sense to keep everything else with the "As" prefix, to indicate that it is accessing the data values.  Could also perhaps use a longer more cumbersome name, but "As" seems reasonable to me.

tensor/datafs/data.go
rcoreilly: see above comments.

tensor/datafs/list.go
rcoreilly: this is Stringer!  here you can see the diff between `AsString` and `String`.

tensor/funcs.go
rcoreilly: type switch is faster though, right?

rcoreilly: but probably not worth all the extra code duplication.

tensor/funcs.go
rcoreilly: yes, called from goal, etc.

tensor/shape.go
rcoreilly: did this throughout on new changes in goal pr.

tensor/funcs.go
kkoreilly: If this is performance critical, then use type switch.

tensor/slice.go
rcoreilly: used step in goal PR

tensor/stats/cluster/cluster.go
rcoreilly: this will be replaced with a datafs representation and I will use full names then.

tensor/stats/cluster/cluster.go
rcoreilly: again, will be datafs as that comment said.

tensor/stats/cluster/cluster.go
rcoreilly: not stringer, and not relevant b/c datafs.

tensor/stats/metric/vec.go
rcoreilly: stands for vectorize, not vector.  not sure I want such a longe filename.

tensor/stats/metric/vec.go
rcoreilly: again, Vectorize.  Vec is fine..

tensor/stats/stats/funcs.go
rcoreilly: there might be other kinds, e.g., NormFunc, within package scope.  doesn't hurt to be specific here I think -- much clearer.

tensor/stats/stats/funcs.go
rcoreilly: better idea?  enum takes precidence.

tensor/stats/stats/README.md
rcoreilly: there are several packages all of the general stats variety.

rcoreilly: but we could discuss whether having a big list at top level might be better?  already kind of a long list -- i"m not sure it is better.

tensor/stats/stats/stats.go
rcoreilly: again, prefer specific names and package might have other functions, e.g., I just moved norm functions in there b/c they fit better.

tensor/stats/stats/vec.go
rcoreilly: "2in" is the concept, not "2In" -- think of it as a single word, with the number providing the upper camel :)

rcoreilly:
see bfff608e7 for some of the fixes from comments just resolved.

tensor/base.go
rcoreilly: Sprint is the best name: printing, to a string.

tensor/datafs/data.go
rcoreilly: per further conversation, we agree about relevant distinctions above.

tensor/funcs.go
rcoreilly: it is the main inner loop for all of numerical computation in tensor land.. so will stick with type switch.  also I don't think we need so many args..

rcoreilly:
more: ec34572cb

tensor/stats/metric/funcs.go
rcoreilly: removed Func in Out64 cases; definitely do call these "Func"s from other packages; cannot export (also MinFunc would conflict with 'min').  I don't see a good alternative, except maybe adding a prefix to the enum that we then strip using enum stripping?  StatMean, StatSum etc?  Maybe that is better?

tensor/funcs.go
rcoreilly: These will be updated to a bitflags b/c need more.

tensor/indexed.go
rcoreilly: yes we do use inside stats/group  renamed to StableSort  etc.

tensor/indexed.go
rcoreilly: fixed in a9fb9164a with an options struct

tensor/indexed.go
rcoreilly: doc'd

tensor/indexed.go
rcoreilly: a9fb9164a struct

tensor/stats/metric/funcs.go
rcoreilly: fixed, enum has Stats* etc.

kkoreilly:
This PR is superseded by #1229, which contains all of these changes and more.

PR #1229, opened 2024-09-17
Goal: Go augmented language

rcoreilly:


rcoreilly:
@kkoreilly really need to fix these timeout tests.. :(

duggavo:
This looks very cool. I love the idea of a simple, cross-platform Go-like language for the GPU.

But wouldn't having a separate repository only for the Go augmented language keep things more clean, both for developers and contributors?

Thank you so much

rcoreilly:
@duggavo thanks for the feedback and good point -- we have discussed this previously and decided that during the intensive development phase we'll keep it in the core monorepo, because it has many dependencies on it, but once everything is more mature and stable we could break it out into a separate repo.

enums/enumgen/methods.go
```diff
@@ -45,8 +45,10 @@ func (i {{.Name}}) {{if .IsBitFlag}} BitIndexString {{else}} String {{end}} () s
 `))
 
 var NConstantTmpl = template.Must(template.New("StringNConstant").Parse(
-	`//{{.Name}}N is the highest valid value for type {{.Name}}, plus one.
+	`//gosl:start
+//{{.Name}}N is the highest valid value for type {{.Name}}, plus one.
 const {{.Name}}N {{.Name}} = {{.MaxValueP1}}
+//gosl:end
```
kkoreilly, line 56: You won't get away with indelibly adding gosl bloat to enumgen.

kkoreilly (changes requested):

enums/enumgen/methods.go
rcoreilly: it is literally just comments. 2 of them. in a generated file. get over it. It is _essential_ because you often need these N values in index computations into enum-based tensor indexing.

kkoreilly (commented):
Wherever gosl and goal end up after this, we should update the links to them in https://github.com/emer/gosl and the initial release blog post respectively.

core/layout.go
```diff
@@ -294,13 +294,13 @@ func (ls *geomState) contentRangeDim(d math32.Dims) (cmin, cmax float32) {
 // totalRect returns Pos.Total -- Size.Actual.Total
 // as an image.Rectangle, e.g., for bounding box
 func (ls *geomState) totalRect() image.Rectangle {
-	return math32.RectFromPosSizeMax(ls.Pos.Total, ls.Size.Actual.Total)
```
kkoreilly, line 298: We had the previous behavior for a reason; I think it might have been related to border, shadow, scrollbar, or parent border radius rendering or something. See c3ca7fa41d79826f21612532fa7524432e35bee8; you must test very thoroughly.

kkoreilly: Actually this is the wrong PR; see #1389.

rcoreilly:
Now at https://github.com/cogentcore/lab -- cogent lab!

PR #1239, opened 2024-09-30
Adding Block and Inline Display styles and Layout logic, for htmlcore.

rcoreilly:


kkoreilly:
This is not the right solution; our existing inline container approach should work, it just needs to be improved.

PR #1270, opened 2024-10-23
Columns

rcoreilly:
This is an option for how columns work in table. @kkoreilly  any thoughts? 

Only real downside is that I had to call the type `Cols` so that the `Columns` field would be visible with the anonymous embed. It is nice that you can now just say `dt.Columns[i]` instead of `dt.Columns.Values[i]`.  But some things do require going through the `dt.Cols` type, but probably not a big deal.

This PR is just to evaluate this particular switch -- will merge back into goal after discussion.


rcoreilly:
Actually, this is a bad idea: forgot that we need to return a Rows view of the column for most cases, so we _don't_ want people directly accessing the Columns[] as such!

PR #1280, opened 2024-10-31
gpudrawer and system.Drawer have SetClearColor 

### LINKED ISSUE #1253: Set underlying WebGPU background color to same as Scene background, so resize doesn't show black regions
### Describe the feature

For the gpu.gpudraw system, the default "clear color" in the renderer should be same as Scene background.  Easy fix.
 

### Relevant code

_No response_

rcoreilly:
that sets the clear color to the color.Scheme.Background so resizing doesn't have bad-looking background.  Fixes #1253

### DIFF
--- a/gpu/gpudraw/drawer.go
+++ b/gpu/gpudraw/drawer.go
@@ -8,6 +8,7 @@ package gpudraw
 
 import (
 	"image"
+	"image/color"
 	"image/draw"
 	"sync"
 
@@ -93,6 +94,14 @@ func (dw *Drawer) Renderer() any {
 		return nil
 	}
 	return dw.System.Renderer
+}
+
+// SetClearColor sets the underlying background color of the Window
+// that is established prior to any image rendering, by the GPU.
+func (dw *Drawer) SetClearColor(clr color.RGBA) {
+	if dw.System != nil {
+		dw.System.SetClearColor(clr)
+	}
 }
 
 // DestSize returns the size of the render destination
--- a/gpu/gpudraw/drawer.go
+++ b/gpu/gpudraw/drawer.go
@@ -8,10 +8,10 @@ package gpudraw
 
 import (
 	"image"
-	"image/color"
 	"image/draw"
 	"sync"
 
+	"cogentcore.org/core/colors"
 	"cogentcore.org/core/gpu"
 )
 
@@ -96,14 +96,6 @@ func (dw *Drawer) Renderer() any {
 	return dw.System.Renderer
 }
 
-// SetClearColor sets the underlying background color of the Window
-// that is established prior to any image rendering, by the GPU.
-func (dw *Drawer) SetClearColor(clr color.RGBA) {
-	if dw.System != nil {
-		dw.System.SetClearColor(clr)
-	}
-}
-
 // DestSize returns the size of the render destination
 func (dw *Drawer) DestSize() image.Point {
 	return dw.System.Renderer.Render().Format.Size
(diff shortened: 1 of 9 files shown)


gpu/gpudraw/drawer.go
```diff
@@ -95,6 +96,14 @@ func (dw *Drawer) Renderer() any {
 	return dw.System.Renderer
 }
 
+// SetClearColor sets the underlying background color of the Window
+// that is established prior to any image rendering, by the GPU.
+func (dw *Drawer) SetClearColor(clr color.RGBA) {
```
kkoreilly: Unless there is any specific reason to the contrary, this should be `c color.Color`, not `clr color.RGBA`. The same applies for all instances of this signature (in the `system.Drawer` interface etc)

system/drawer.go
```diff
@@ -157,3 +162,6 @@ func (dw *DrawerBase) Renderer() any {
 	// no-op
 	return nil
 }
+
+func (dw *DrawerBase) SetClearColor(clr color.RGBA) {
+}
```
kkoreilly: Prefer to keep empty no-op functions on one line.

system/driver/base/window_multi.go
```diff
@@ -105,6 +106,7 @@ func (w *WindowMulti[A, D]) SetLogicalDPI(dpi float32) {
 	w.Mu.Lock()
 	defer w.Mu.Unlock()
 	w.LogDPI = dpi
+	w.Draw.SetClearColor(colors.ToUniform(colors.Scheme.Background))
```
kkoreilly: Why are you doing this in `window_multi.go` and everything else in platform-specific files? In general, if there is any way to put this in one centralized place, that would be great. (Or at least put a helper function somewhere that you call that does this).

kkoreilly: Just to be clear, windows get resized and correct background colors are needed on mobile and web as well, and those do not go through `WindowMulti`, so your hack for `SetLogicalDPI` updating the color after a scheme change isn't really tenable.

kkoreilly: What about making `gpudraw.NewDrawer` automatically do this?

kkoreilly (changes requested):

system/drawer.go
rcoreilly: note that this is not the case with the code just above this.

rcoreilly: has explicit / redundant no-op comments.

system/driver/base/window_multi.go
rcoreilly: newdrawer can set it initially but it doesn't know when it changes.

rcoreilly: actually it looks like this setting doesn't incur any cost so we could just update it every time.

rcoreilly: when do windows get resized on mobile?

system/drawer.go
kkoreilly: You can either do nothing or a no-op comment and newline, but you can't do a newline without a no-op comment, since that is pointless.

system/driver/base/window_multi.go
kkoreilly: It is irrelevant when windows get resized on mobile if you just set the clear color in `Start` as you do with your new approach.

kkoreilly (approved):
You can definitely see that the new approach is much better; any time you find yourself adding things for each platform separately, you should think about DRY.

PR #1281, opened 2024-10-31
fix for slow table updating with many rows -- UpdateMaxWidths is the culprit.

### LINKED ISSUE #1128: Table performance with 100,000 rows
### Describe the feature

I have been experimenting with Cogent Core for our needs and have encountered possible performance issues which would be useful if addressed. Our requirements deal with large amounts of data.

For example, create a table with two fields and 100,000 rows and scroll the table, the redraw is laggy. Now, I know 100,000 rows is a lot, but just wondering, as the API stabilises - is now is the time to address and improve performance across all widgets?

An interim solution is to filter the data before adding to the table which is probably what we'll do, but I was interested in the thoughts of the team regarding profiling and improving performance and whether this is planned?

### Relevant code

_No response_

rcoreilly:
Fixes #1128

### DIFF
--- a/core/list.go
+++ b/core/list.go
@@ -255,8 +255,6 @@ func (lb *ListBase) Init() {
 		// absorb horizontal here, vertical in view
 		s.Overflow.X = styles.OverflowAuto
 		s.Grow.Set(1, 1)
-		ls := lb.This.(Lister)
-		ls.UpdateMaxWidths()
 	})
 	if !lb.IsReadOnly() {
 		lb.On(events.DragStart, func(e events.Event) {
@@ -401,6 +399,7 @@ func (lb *ListBase) SetSliceBase() {
 		lb.SelectedIndex = -1
 	}
 	lb.ResetSelectedIndexes()
+	lb.This.(Lister).UpdateMaxWidths()
 }
 
 // SetSlice sets the source slice that we are viewing.
@@ -425,11 +424,11 @@ func (lb *ListBase) SetSlice(sl any) *ListBase {
 		return lb
 	}
 
-	lb.SetSliceBase()
 	lb.Slice = sl
 	lb.sliceUnderlying = reflectx.Underlying(reflect.ValueOf(lb.Slice))
 	lb.isArray = reflectx.NonPointerType(reflect.TypeOf(sl)).Kind() == reflect.Array
 	lb.elementValue = reflectx.SliceElementValue(sl)
+	lb.SetSliceBase()
 	return lb
 }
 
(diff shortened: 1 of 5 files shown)


core/list.go
```diff
@@ -255,6 +255,8 @@ func (lb *ListBase) Init() {
 		// absorb horizontal here, vertical in view
 		s.Overflow.X = styles.OverflowAuto
 		s.Grow.Set(1, 1)
+		svi := lb.This.(Lister)
```
kkoreilly: `svi` is not an appropriate name for a `Lister`, even if you copied it from some other place; always correct these old names when you come across them. `ls` is standard.

core/list.go
```diff
@@ -255,6 +255,8 @@ func (lb *ListBase) Init() {
...
+		svi.UpdateMaxWidths()
```
kkoreilly: Can you give a coherent explanation of why this needs to go in `Styler` and not `Updater`?

examples/demo/demo.go
```diff
@@ -389,12 +389,21 @@ func collections(ts *core.Tabs) {
...
 
-	tbl := make([]*tableStruct, 50)
+	// tbl := make([]*tableStruct, 50)
+	// for i := range tbl {
+	// 	ts := &tableStruct{Age: i, Score: float32(i) / 10}
+	// 	tbl[i] = ts
+	// }
+	// tbl[0].Name = "this is a particularly long field"
+	// ttab, _ := vts.NewTab("Tables")
+	// core.NewTable(ttab).SetSlice(&tbl)
```
kkoreilly: What is your plan for this commented out code?

examples/demo/demo.go
```diff
@@ -528,6 +537,11 @@ func (ts *testStruct) ShouldDisplay(field string) bool {
 	return true
 }
 
+type smallStruct struct { //types:add
```
kkoreilly: Why `smallStruct`? What is wrong with `tableStruct`?

kkoreilly (changes requested):

core/list.go
rcoreilly: styler is only run when there are more significant changes; Update is run to redraw during scrolling given the nature of the way this widget works, with static widgets representing different content as it scrolls.

core/list.go
rcoreilly: fixed

examples/demo/demo.go
rcoreilly: good catch -- forgot to undo

core/list.go
kkoreilly: That is a complete inversion of normal expectations; it is the case for every other widget that `Update` is a much more major process and `Style` is run more frequently. Are you re-running the `Maker` while you scroll? I thought the whole point was to reuse the same widgets for the table. It seems like there should be a better mechanism than `UpdateWidget` for updating while scrolling.

rcoreilly: I can never remember what happens when.. Update does run Style, and in fact the slow UpdateMaxWidths call is being made in the Style. I'm looking into it.

rcoreilly: ok now calling UpdateWidths precisely when needed.  tested to 10_000_000 rows with no problems!

kkoreilly (approved):

PR #1292, opened 2024-11-10
Add manual install commands for Linux to install docs

### LINKED ISSUE #1260: Cannot try cogentcore due to su password requirement
### Describe the feature

When I tried `core setup` it would not let me proceed without me entering my su password. I am not willing to do that since I don't know what it would do to my system.

I always install Go locally, e.g., in `~/opt/go` with my packages in `~/opt/gopkg` so anything installed into Go doesn't need su access.

And as for third-party libs, e.g., OpenGL or whatever cogentcore uses, a simple list of what needs installing would allow me to do it myself.

PS having an executable called `core` on Unix does not make any sense; I suggest using a different name, `ccore` or pretty well anything else.

### Relevant code

_No response_

### LINKED ISSUE #1275: Manual Install Documentation
### Describe the feature

Hey there, this sure looks like an interesting project.  Looking forward to trying it out.  I was a little concerned about the need for sudo on the core setup command recommended here https://www.cogentcore.org/core/setup/install though.

In general a Linux user is going to be a little cautious about what they let run with sudo.  I was able to track down the setup file and see that it's issuing package commands so it makes sense that you'd need sudo for that.  But I'd still rather handle dependencies myself and still know that I'm setting myself up for success.  

I'd imagine that this would look like a page of documentation that just says "these are the packages we need" for the distros you support.  This would also help anyone running a distro that you don't support satisfy those dependencies (assuming there aren't other issues that they may encounter) by giving them a list to work from to look up what the equivalent is for them.

### Relevant code

_No response_

kkoreilly:
Fixes #1275 and fixes #1260.

### DIFF
--- a/docs/content/2-setup/0-install.md
+++ b/docs/content/2-setup/0-install.md
@@ -3,8 +3,23 @@ Note: we recommend you read the [basics](../basics) and [tutorials](../tutorials
 1. Download and install Go from [the Go website](https://go.dev/doc/install) if you do not already have Go 1.22+ installed.
 2. Run `go install cogentcore.org/core/cmd/core@main` to install the Cogent Core command line tool. You should run this command again each time you update to a new version of Cogent Core.
 3. Run `core setup` to install platform-specific dependencies. Please first read all of the information below:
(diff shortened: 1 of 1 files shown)


kkoreilly:
This is superseded by #1321, which added the manual Linux install commands to the install page.

PR #1300, opened 2024-11-12
Add generic Shift+ nav key selection logic

rcoreilly:
And update the Control+ and Alt+ modifiers for Home, End so they go to DocHome, DocEnd consistently across platforms.

Here are the new key map tables: https://github.com/cogentcore/core/discussions/650

Fixes #1282


core/textfield.go
```diff
@@ -149,6 +149,9 @@ type TextField struct { //core:embedder
 	// selectMode is whether to select text as the cursor moves.
 	selectMode bool
 
+	// selectModeShift is whether selectmode was turned on because of the shift key.
+	selectModeShift bool
```
kkoreilly, line 153: We don't also need this for text editor?

keymap/keymap.go
```diff
@@ -42,6 +42,7 @@ const (
 	DocEnd  // end-of-doc Control / Alt / Shift +End
 	WordRight
 	WordLeft
+	// WordLeft is the final navigation function -- all above also allow Shift+ for selection.
 	FocusNext // Tab
```
kkoreilly, line 47: Add a line between the comment and next enum here to avoid erroneous association?

keymap/keymap.go
```diff
@@ -152,7 +153,20 @@ func SetActiveMapName(mapnm MapName) {
...
-	return (*ActiveMap)[chord]
+	f, ok := (*ActiveMap)[chord]
+	if ok {
+		return f
+	}
+	if strings.Contains(string(chord), "Shift+") {
+		nsc := key.Chord(strings.ReplaceAll(string(chord), "Shift+", ""))
+		f, ok = (*ActiveMap)[nsc]
+		if ok {
+			if f <= WordLeft { // automatically allow +Shift for nav
```
kkoreilly: Combine these two if statements with `&&`?

kkoreilly (changes requested):

core/textfield.go
rcoreilly: nope -- it works fine as-is -- has more sophisticated logic.

kkoreilly (changes requested):
Overall looks good and shortcuts all work for me, except for Cmd+Backspace in text field and text editor on macOS; can you add support for that?

keymap/standard.go
```diff
@@ -60,6 +60,7 @@ var StandardMaps = Maps{
 		"Backspace":           Backspace,
 		"Control+Backspace":   BackspaceWord,
 		"Alt+Backspace":       BackspaceWord,
+		"Meta+Backspace":      BackspaceWord,
```
kkoreilly, line 63: Cmd+Backspace is supposed to delete to the start of the line, not the start of the word. Also, the same is true for Cmd+Delete.

kkoreilly (changes requested):

keymap/standard.go
kkoreilly: Actually, it looks like Cmd+Delete isn't generally supported, so we just need Cmd+Backspace to work for delete line.

rcoreilly: We do not have a Function for delete to start of line. You can file an issue for that, but it is not going to happen in this PR. 

Also, clarify Cmd+Delete: is that the same as delete-to-end-of-line?  That would be more consistent with "forward delete" vs. "backward delete". We DO already have delete-to-end-of line: Kill -- that is easy.

rcoreilly: oh right you said no Cmd+Delete already?  where id that comment go?

kkoreilly: Theoretically:

* Cmd+Backspace: delete to start of line
* Cmd+Delete: delete to end of line

In practice, it seems like other apps support Cmd+Backspace but not Cmd+Delete. I use Cmd+Backspace a lot. There is no existing keymap for the inverse of Kill?

rcoreilly: nope. again, file an issue and approve this. and LMK if you want to go ahead and map Cmd+Delete to Kill.

Also for cross-platform consistency, what is the non-Meta (Cmd) version of Cmd+Delete = Kill for other platforms? Alt? Control?  which one goes to word and which one goes to line?

looking at: https://github.com/cogentcore/core/discussions/650 -- it seems like Control+ is word in general. So it would overall be more consistent to have Control+nav = word; Alt+nav = line. and on mac, Meta = Cmd works just like Control.

again probably better to sort this out in another issue, and go ahead and merge this.

kkoreilly (approved):

PR #1306, opened 2024-11-14
 A set of stage / window sizing fixes

### LINKED ISSUE #1298: Support a growing window
### Describe the bug

Providing GrowWrap = true to a new body makes it unable to display any content. Settings are taken directly from https://www.cogentcore.org/core/advanced/styling/.

Expected behavior: the windows contains a text widget with "Picker" and a picker with options "foo" and "bar". The window wraps tightly around contents of the child frame.

Result: a black window appears with a 0,0 box 
result

### How to reproduce

`core run` the code in the issue

### Example code

```go
package main

import (
	"fmt"

	"cogentcore.org/core/core"
	"cogentcore.org/core/events"
	"cogentcore.org/core/styles"
)

func main() {
	b := core.NewBody()
	b.Styler(func(s *styles.Style) {
		s.GrowWrap = true
		s.Gap.X = units.Value{Value: 0.5, Unit: units.UnitEm}
		s.Gap.Y = units.Value{Value: 0.5, Unit: units.UnitEm}
	})

	fr := core.NewFrame(b)
	fr.Styler(func(s *styles.Style) {
		s.Align = styles.AlignSet{Items: styles.Center}
	})

	t := core.NewText(fr).SetText("Picker")
	t.Styler(func(s *styles.Style) {
		s.SetTextWrap(false)
	})
	ch := core.NewChooser(fr).SetStrings("foo", "bar")
	ch.OnChange(func(e events.Event) {
		core.MessageSnackbar(fr, fmt.Sprintf("Selected %v", ch.CurrentItem.Value))
	})

	b.RunMainWindow()
}
```

### Relevant output

_No response_

### Platform

Windows

rcoreilly:
* change the popup sizing and positioning logic from c88146289 to only apply when popup is a significant fraction of the size of the source window that it is a popup of. much better.

* add an option on Stage to disable min size of a new window, and also to specify the size of the window manually. 

* add a `ResizeToContents` function to Scene that resizes an existing Scene to fit current contents, and rename `prefSize` -> `contentSize` 

fixes #1283 and fixes #1298


rcoreilly:
@kkoreilly can you test resize to contents?

core/render.go
```diff
@@ -255,20 +255,20 @@ func (sc *Scene) doRebuild() {
 	sc.layoutRenderScene()
 }
 
-// prefSize computes the preferred size of the scene based on current contents.
+// contentsSize computes the size of the scene based on current contents.
 // initSz is the initial size -- e.g., size of screen.
 // Used for auto-sizing windows.
-func (sc *Scene) prefSize(initSz image.Point) image.Point {
+func (sc *Scene) contentsSize(initSz image.Point) image.Point {
```
kkoreilly: `sizeToContent` is probably a better name.

core/scene.go
```diff
@@ -124,11 +124,11 @@ const (
 	// copy the image up to the GPU or not.
 	sceneImageUpdated
 
-	// scenePrefSizing means that this scene is currently doing a
+	// sceneContentsSizing means that this scene is currently doing a
 	// PrefSize computation to compute the size of the scene
```
kkoreilly: not PrefSize

core/scene.go
```diff
@@ -124,11 +124,11 @@ const (
...
 	// (for sizing window for example); affects layout size computation
 	// only for Over
-	scenePrefSizing
+	sceneContentsSizing
```
kkoreilly: `sceneSizingToContent`?

core/scene.go
```diff
@@ -293,6 +293,26 @@ func (sc *Scene) resize(geom math32.Geom2DInt) {
 	sc.NeedsLayout()
 }
 
+// ResizeToContents resizes the scene so it fits the current contents.
+// Only applicable to Desktop systems where windows can be resized.
+// Extra size is added to the amount computed to hold the contents,
+// which is needed in cases with wrapped text elements, which don't
+// always size accurately.
+func (sc *Scene) ResizeToContents(extra image.Point) {
```
kkoreilly: `ResizeToContent`? Or will people get confused with #1304? Making the user pass `extra` is also not good.

core/stage.go
```diff
@@ -120,6 +120,10 @@ type Stage struct { //types:add -setters
 	// [WindowStage]s take up the entire window they are created in.
 	FullWindow bool
 
+	// For NewWindow stages, use the actual contents size, instead of
+	// enforcing a minimum size as a function of the total screen size.
+	ContentsSize bool
```
kkoreilly: `SizeToContent`?

examples/demo/demo.go
```diff
@@ -597,6 +598,12 @@ func dialogs(ts *core.Tabs) {
 		d.RunWindowDialog(nd)
 	})
 
+	rw := core.NewButton(drow).SetText("Resize To Contents")
```
kkoreilly: This doesn't really belong here, and it should be sentence case.

examples/demo/demo.go
```diff
@@ -597,6 +598,12 @@ func dialogs(ts *core.Tabs) {
...
+	rw.SetTooltip("resizes this window to fit the current contents")
```
kkoreilly: Tooltips should start capitalized.

system/driver/desktop/window.go
```diff
@@ -223,6 +223,12 @@ func (w *Window) SetWinSize(sz image.Point) {
 	})
 }
 
+func (w *Window) SetSize(sz image.Point) {
```
kkoreilly, line 226: `SetSize` and `SetWinSize` both here and in `WindowMulti` should set `w.PixSize` and `w.WnSize` respectively.

kkoreilly (changes requested):
Definitely somewhat buggy for me. Also centralized docs would be nice...

core/render.go
rcoreilly: this function does _not_ change the size. it returns the size that fits the content. sizeToContent sounds like it changes the size. sizeOfContent would be more accurate, but contentsSize is the same thing and simpler.

core/scene.go
rcoreilly: nope, see above.

examples/demo/demo.go
rcoreilly: is there a better place?

system/driver/desktop/window.go
rcoreilly: I believe those get set when the window resize event comes back through.

core/render.go
kkoreilly: Why not `contentSize`? I don't really see what the `s` is doing; is it "contents size" or "content's size"? Isn't "content" already plural?

examples/demo/demo.go
kkoreilly: On the same page either in "Windows" or in its own section?

core/scene.go
kkoreilly: `sceneContentSizing` implies that it is sizing the content, which is not true.

rcoreilly: that is precisely what it is doing.

core/scene.go
rcoreilly: made extra optional

core/stage.go
rcoreilly: ContentSize seems more consistent.

core/stage.go
```diff
@@ -120,9 +120,9 @@ type Stage struct { //types:add -setters
 	// [WindowStage]s take up the entire window they are created in.
 	FullWindow bool
 
-	// For NewWindow stages, use the actual contents size, instead of
+	// For NewWindow stages, use the actual content size, instead of
 	// enforcing a minimum size as a function of the total screen size.
-	ContentsSize bool
+	ContentSize bool
```
kkoreilly: I can see your point for the other things that they return the content size, but this is about whether you should size to content. `ContentSize bool` doesn't really make sense; I much prefer `SizeToContent`.

core/scene.go
```diff
@@ -293,12 +292,12 @@ func (sc *Scene) resize(geom math32.Geom2DInt) {
...
 
-// ResizeToContents resizes the scene so it fits the current contents.
+// ResizeToContent resizes the scene so it fits the current content.
 // Only applicable to Desktop systems where windows can be resized.
-// Extra size is added to the amount computed to hold the contents,
+// Optional extra size is added to the amount computed to hold the contents,
 // which is needed in cases with wrapped text elements, which don't
 // always size accurately.
-func (sc *Scene) ResizeToContents(extra image.Point) {
+func (sc *Scene) ResizeToContent(extra ...image.Point) {
```
kkoreilly, line 300: I don't want optional extra; if it is needed, make it irrevocably added for now with a TODO to remove it later once we fix #1307.

examples/demo/demo.go
```diff
@@ -642,6 +636,12 @@ func dialogs(ts *core.Tabs) {
 		core.NewText(d).SetType(core.TextSupporting).SetText("A standalone window that opens in the same system window")
 		d.NewWindow().SetNewWindow(false).SetDisplayTitle(true).Run()
 	})
+
+	rw := core.NewButton(wrow).SetText("Resize to content")
+	rw.SetTooltip("Resizes this window to fit the current content")
+	rw.OnClick(func(e events.Event) {
+		wrow.Scene.ResizeToContent(image.Point{0, 40}) // note: size is not correct due to wrapping? #1307
```
kkoreilly: Again, this is not the place to put the extra and the note.

kkoreilly (changes requested):

core/scene.go
rcoreilly: that is a less future-compatible strategy, once we remove it. if it works without then shouldn't pass anything.

examples/demo/demo.go
rcoreilly: ?? where else?

rcoreilly: image.Pt(0,40)

core/stage.go
rcoreilly: New plan: `UseMinSize`  -- set to true by default.  also apply to dialogs.

kkoreilly (approved):

PR #1337, opened 2024-11-25
Added Deferred functions, set by Defer method, which are called after…

### LINKED ISSUE #1194: Need an event for when a tab newly becomes visible, e.g., Show
### Describe the feature

@kkoreilly we have discussed this but I don't remember the current status.  In the emer sims, we really need to be able to get the current state of the Network when someone switches back to that tab.  Definitely do not want tab updating when not visible as it is expensive.

It would make sense for this to be the Show event, although there was some issue about whether Show is a "once" only event or not (could always use `sync.Once` to ensure where needed, presumably).

### Relevant code

_No response_

### LINKED ISSUE #1322: Switching Pages Breaks Widget Focus
### Describe the bug

The new page switching is really nice, unfortunately it doesn't seem to play completely nicely with focus.  This may also be user error.

### How to reproduce

I ran this two ways to try and understand what was going on, first:

- Run example code (slightly modified from tutorial)
- Observe that the text box is focused as expected and styled to indicate that
- Enter a string and click sign in button
- Observe that the text box on the "home" page is not styled as focused
- Attempt to type some characters slowly one at a time
- Observe that the text box isn't actually focused for the first few key presses
- Click sign out button
- Observe that the text box is not styled as focused this time
- Attempt to type some characters slowly one at a time
- Observe that the text box isn't actually focused for the first few key presses

Second, swap to using SetFocusEvent instead by switching which lines are commented and then:

- Run example code
- Observe that the text box is not styled as focused
- Attempt to enter some characters
- Observe that text box is actually focused
- Click sign in button
- Observe that text box on "home" page is not styled as focused
- Attempt to type some characters
- Observe that the text box is actually focused
- Click sign out button
- Attempt to enter text once again, observing that once again the text box is actually focused but not styled as such

### Example code

```go
package main

import (
	"cogentcore.org/core/core"
	"cogentcore.org/core/events"
)

type user struct {
	Username string
}

func main() {
	b := core.NewBody()
	u := &user{}
	pg := core.NewPages(b)
	pg.AddPage("sign-in", func(pg *core.Pages) {
		userField := core.NewTextField(pg)
		userField.StartFocus()
		//userField.SetFocusEvent()
		core.NewButton(pg).SetText("Sign in").OnClick(func(e events.Event) {
			u.Username = userField.Text()
			pg.Open("home")
		})
	})
	pg.AddPage("home", func(pg *core.Pages) {
		core.NewText(pg).SetText("Welcome, " + u.Username + "!").SetType(core.TextHeadlineSmall)
		newField := core.NewTextField(pg)
		newField.StartFocus()
		//newField.SetFocusEvent()
		core.NewButton(pg).SetText("Sign out").OnClick(func(e events.Event) {
			*u = user{}
			pg.Open("sign-in")
		})
	})

	b.RunMainWindow()
}
```

### Relevant output

_No response_

### Platform

Linux

rcoreilly:
… a full scene update. Tabs and Pages use to set StartFocus and call OnShow on newly shown elements. Fixes #1322 Fixes #1194


core/render.go
```diff
@@ -398,8 +406,50 @@ func (wb *WidgetBase) renderChildren() {
 	})
 }
 
-////////////////////////////////////////////////////////////////////////////////
-//  Standard Box Model rendering
+////////  Defer
+
+// Defer adds a functions that will be called after a full [Scene] update,
```
kkoreilly: `a function`.

core/render.go
```diff
@@ -398,8 +406,50 @@ func (wb *WidgetBase) renderChildren() {
...
+// including on the initial Scene render.
+// At this point event sending functions will work as expected, etc.
```
kkoreilly: `In this function event sending...`

core/render.go
```diff
@@ -398,8 +406,50 @@ func (wb *WidgetBase) renderChildren() {
 	})
 }
 
-////////////////////////////////////////////////////////////////////////////////
-//  Standard Box Model rendering
+////////  Defer
+
+// Defer adds a functions that will be called after a full [Scene] update,
+// including on the initial Scene render.
```
kkoreilly: `After the function is called, it is removed`.

core/render.go
```diff
@@ -398,8 +406,50 @@ func (wb *WidgetBase) renderChildren() {
...
+	wb.Deferred = append(wb.Deferred, fun)
+	if wb.Scene != nil {
+		wb.Scene.setFlag(true, sceneHasDeferred)
+	}
+}
+
+// runDeferred runs deferred functions on all widgets in the scene.
+func (sc *Scene) runDeferred() {
+	sc.WidgetWalkDown(func(cw Widget, cwb *WidgetBase) bool {
+		if cwb.Deferred != nil {
```
kkoreilly: Pointless nil check.

core/render.go
```diff
@@ -398,8 +406,50 @@ func (wb *WidgetBase) renderChildren() {
...
+				f()
+			}
+			cwb.Deferred = nil
+		}
+		return tree.Continue
+	})
+}
+
+// DeferShown adds a [Widget.Deferred] function to activate [WidgetBase.StartFocus]
+// and call [WidgetBase.Shown]. This is called in [Tabs] and [Pages] for example, when
```
kkoreilly: Put Shown and StartFocus in the order they are called (ie: swap them).

core/render.go
```diff
@@ -445,7 +494,7 @@ func (wb *WidgetBase) winPos(x, y float32) image.Point {
 	return pt
 }
 
-// Profiling and Benchmarking, controlled by settings app bar:
+//////// Profiling and Benchmarking, controlled by settings app bar:
```
kkoreilly: It interpreted this as a comment directive... (see in typegen below)

core/widgetevents.go
```diff
@@ -467,14 +467,14 @@ func (wb *WidgetBase) HandleClickOnEnterSpace() {
...
+////////	Focus
 
 // SetFocus sets the keyboard input focus on this item or the first item within it
 // that can be focused (if none, then just sets focus to this object).
 // This does not send an [events.Focus] event, which typically results in
 // the widget being styled as focused. See [WidgetBase.SetFocusEvent] for
-// a version that does.
+// a version that does. Also see [WidgetBase.StartFocus] which must be
+// called instead during initial construction prior to the initial scene render.
 func (wb *WidgetBase) SetFocus() {
```
kkoreilly: Would any end user call `SetFocus` instead of `SetFocusEvent`? Should this be unexported and `SetFocusEvent` get a better name?

events/types.go
```diff
@@ -214,10 +214,11 @@ const (
 	Input
 
 	// Show is sent to widgets when their Scene is first shown to the user
-	// in its final form. Listening to this event enables widgets to perform
-	// initial one-time activities on startup, in the context of a fully
-	// rendered display. This is guaranteed to only happen once per widget
-	// per Scene.
+	// in its final form, and whenever a major content managing widget
+	// (e.g., [core.Tabs], [core.Pages]) shows a new element.
```
kkoreilly: We should probably mention the actual `core.WidgetBase.Shown` function that does that.

core/render.go
```diff
@@ -398,8 +406,50 @@ func (wb *WidgetBase) renderChildren() {
...
+// an element is newly shown, so that these elements can perform OnShow updating as needed.
+func (wb *WidgetBase) DeferShown() {
+	wb.Defer(func() {
+		wb.Shown()
+		wb.Scene.Events.activateStartFocus()
+	})
+}
+
+// Shown sends [events.Show] to all widgets from this one down.
+func (wb *WidgetBase) Shown() {
```
kkoreilly, line 435: Mention `DeferShown`.

kkoreilly (changes requested):

core/widgetevents.go
kkoreilly: Going with `SetFocusQuiet` instead of `SetFocus` and `SetFocus` instead of `SetFocusEvent`, since quiet is rarer and thus should be the one with the suffix.

kkoreilly (approved):

PR #1346, opened 2024-11-25
Add support for custom linker flags in core tool, add trimpath and windowsgui, and use the directory name for binary names on desktop

### LINKED ISSUE #1345: Add support for custom linker flags in core tool and support windowsgui option
### Describe the feature

As discussed in #1294, we should add support for custom linker flags in the core tool. We should also add support for `-H=windowsgui` (on by default for `core build`, `core pack`, etc, off by default for `core run`). 

### Relevant code

_No response_

kkoreilly:
This PR adds support for custom `ldflags` in the `core` tool, adds `trimpath` and `windowsgui` options that are on by default for non-`core run` commands, and updates the default executable name on desktop platforms to be the directory name instead of the app name, as that is more standard and has no spaces; the user-friendly name is still used for packaging, but is not needed for development executables.

Fixes #1345


rcoreilly (approved):
thank you so much for fixing the binary name for desktop non-package builds. I greatly appreciate your consideration in this matter.

PR #1347, opened 2024-11-25
Use GitHub Artifacts to upload and download testdata images in CI

kkoreilly:
This allows for our unit tests to actually compare current images to previous ones, ensuring proper regression testing; if the image result of any unit test changes, CI will fail and we will have to fix the issue or update the images using `-tags update` if it is intended. I added a manual button trigger for the workflow that enables `-tags update` so we don't have to manually change the workflow back and forth every time there are changes to the images.

This PR also makes offscreen a multi-window platform and merges AssertRender and AssertRenderScreen to fix testing unreliability for multi-scene tests.


kkoreilly:
This is all working now except for a number of random variations in rendering in several tests, resulting in CI failures.

kkoreilly (commented):
This all seems to be finally working!

rcoreilly (dismissed):

rcoreilly (approved):
Again, no annoying comments :)

PR #1341, opened 2024-11-25
significantly improved support for multi-monitor setups on desktop; other screen, stage, scene improvements

rcoreilly:
Add screen name to saved window geometry so it prefers to open there, and add option for screen name in opening new windows, in stage and lower level system api. also system.Window.Position() always returns screen-relative position. all working as expected.

Picks up from progress in #1339

* Fixes #603 -- always uses current screen for refresh (UpdateAll).
* Fixes #997 and Fixes #1140 -- can do e.g., `b.NewWindow().SetScreen(1).RunMain()` to specify screen 1 as the default to open a new window.
* Fixes #509 -- added ConstrainToFrame that keeps full window frame in view.
* Fixes #1329 -- added Resizable flag on Stage
* Fixes #1357 -- detects monitor changes after waking from sleep.
* Fixes #1359 -- window Screen() is not always up to date at monitor change events: use wingeom version.
* Fixes #1083 -- screen-specific zoom now used for zoom shortcut and made read-only and fixed comments.
* Fixes #947 -- never saw this in all my testing of plugging and unplugging, but pushed an extra sanity fix anyway.
* Fixes #1360 -- fixed issues in the ConstrainFrame code and detect the -32000 special windows position value.
* May help with #1319 and #1247 based on updates in device config.

PR #1364, opened 2024-12-09
Stage.Screen defaults to -1, is only used when set (0 or higher); other screen fixes

rcoreilly:
A last minute change in PR #1341 ended up overriding the effects of Stage.Screen!  This fixes that issue and documents the new logic: 

	// Screen specifies the screen number on which a new window is opened
	// by default on desktop platforms. The default behavior (Screen = -1)
	// will open the first window on screen 0 (default primary screen) and
	// any subsequent windows will open on the same screen as the currently
	// active window. In any case, the automatically saved last position for
	// a window (by window title) always has the final say if it exists.
	// Use [TheApp].ScreenByName("name").ScreenNumber to get the number by name.

This is implemented by the `targetScreen` method, used consistently now.

Also added `WindowGeometryDisabled` flag to allow testing of code like this, without interference from saved prefs.

@kkoreilly I left demo with the testing code so you can give it a try -- be sure to remove before merging!


core/mainstage.go
```diff
@@ -178,7 +178,13 @@ func (st *Stage) firstWindowStages() *stages {
 	return ms
 }
 
-func (st *Stage) currentScreen() *system.Screen {
+// targetScreen returns the screen to use for opening a new window
+// based on Screen field, currentRenderWindow's screen, and a fallback
+// default of Screen 0.
+func (st *Stage) targetScreen() *system.Screen {
+	if st.Screen >= 0 && st.Screen < system.TheApp.NScreens() {
+		return system.TheApp.Screen(st.Screen)
```
kkoreilly: Use `TheApp` (in core) instead of manually referencing system.

core/windowgeometry.go
```diff
@@ -21,8 +21,14 @@ import (
...
 
-// theWindowGeometrySaver is the manager of window geometry settings
-var theWindowGeometrySaver = windowGeometrySaver{}
+var (
+	// theWindowGeometrySaver is the manager of window geometry settings
+	theWindowGeometrySaver = windowGeometrySaver{}
+
+	// WindowGeometryDisabled disables the saving and loading of window geometry
+	// data, for testing purposes.
+	WindowGeometryDisabled = false
```
kkoreilly: Put in `DebugSettings`

kkoreilly (changes requested):

core/windowgeometry.go
kkoreilly: Also name should be `DisableWindowGeometry`.

kkoreilly (approved):

PR #1367, opened 2024-12-12
Compute DPI as 160 * DPR * Scale instead of using Physical DPI

### LINKED ISSUE #1358: Apps look too small on some monitors
### Describe the bug

On some monitors, app content looks too small by default. This is true on one of my secondary monitors, and other people have also reported [similar things](https://gophers.slack.com/archives/C07ENRTB2F7/p1728243094303549).

### How to reproduce

It is not really clear what causes this; it seems like it only happens on 1080p non-HiDPI monitors. It has happened on macOS, Windows, and Linux. A temporary workaround is to increase the zoom level using `Ctrl+` or in the settings.

### Example code

_No response_

### Relevant output

_No response_

### Platform

macOS

kkoreilly:
This PR makes `logical dpi = 160 * dpr * scale`, as that is consistent with all other apps and the web. The previous behavior of having the physical size of widgets be the same on all platforms sounded great in theory but resulted in low quality font rendering and inconsistency with all other apps. This new paradigm looks completely correct on all of my monitors and VMs, whereas the old one required manually zooming in on VMs and low-DPI monitors.

This PR also hides wgpu warnings since there are a lot of extraneous ones, and it updates our version of webgpu to include https://github.com/cogentcore/webgpu/commit/ad7475f3b4dde476d971b4aca10a12c160dd8dfb, which is a key fix for a `Cannot perform Construct on a detached ArrayBuffer` crash on web. It also fixes `GLFWCreateWindow` not building on web.

Fixes #1358 (the issue was the incorrect logical DPI computation for low-DPI monitors as discussed above)


rcoreilly (changes requested):
This will not be merged without an option to use physical size.

rcoreilly (approved):
After further discussion, we can use the Physical DPI field to accomplish exact print-size rendering of things like print preview as needed in future apps, and the fact that this convention is used in all other apps makes sense for now.

PR #1372, opened 2024-12-15
Improve cursor sizes, colors, and shadows

### LINKED ISSUE #662: Add support for cursor shadows
At least need shadows, so it stands out in all cases.

The disabled cursor also is effectively invisible -- specifically bad case of this.

### LINKED ISSUE #1119: Change cursor size based on zoom / device pixel ratio
### Describe the bug

On Windows 11 running the demo the mouse cursor size is different to the usual size for the rest of the system while inside the Cogent app.

Large cursor when hovering over the title bar:
IMG_8025

Changes to a small size when inside the window:
IMG_8028

### How to reproduce

go run cogentcore.org/core/examples/demo@main

This happens the "Scale" setting is set to 200% in Windows Display Settings.  When I change the scale to 100% the mouse cursor is the same inside Cogent apps as everywhere else.

### Example code

_No response_

### Relevant output

_No response_

### Platform

Windows

kkoreilly:
This PR makes cursor sizes correct by taking into account DPI. It also makes cursor colors automatically switch based on the color scheme to look good. In addition to various other cleanups and rendering fixes, it also adds shadow rendering to make cursors stand out more.

Fixes #1119 and fixes #662


rcoreilly (approved):
nice!

PR #1376, opened 2024-12-16
Use debug.ReadBuildInfo for version displaying in core tool (cli) and as a backup in apps

### LINKED ISSUE #1370: Binary Versioning
### Describe the feature

I notice that the `core` binary does not have the version compiled into it.

This begs the question: when did I build this binary and is this the latest version? There is no way to tell.

So I propose a simple method which I've recently devised to compile the version into the binary.

### Relevant code

it's possible to get the informaton regarding the latest commits to a branch of this repo (for example `main`) with `go list` ; example:

```
$ go list -m -json github.com/cogentcore/core@main

{
	"Path": "github.com/cogentcore/core",
	"Version": "v0.3.8-0.20241213221857-7f699071cef2",
	"Query": "main",
	"Time": "2024-12-13T22:18:57Z",
	"GoMod": "/home/user/go/pkg/mod/cache/download/github.com/cogentcore/core/@v/v0.3.8-0.20241213221857-7f699071cef2.mod",
	"GoVersion": "1.22",
	"Origin": {
		"VCS": "git",
		"URL": "https://github.com/cogentcore/core",
		"Hash": "7f699071cef2f7365f35c8b470e4066a8ce3b0cc"
	}
}

```

The above should be executed in a subshell to set a variable at compile time via `-ldflags= -X ...` - in a library to be created. This can be, for instance pkg/buildinfo or, I might even suggest putting a file like `core.go` in the root of this repository where that can be set.

So, for instance, instead of:

```
go install cogentcore.org/cmd/core@main
```

one might run

```
go install -ldflags=" -X github.com/cogentcore/core.buildInfo=$(go list -m -json github.com/cogentcore/core@main)" cogentcore.org/cmd/core@main

```

and there should perhaps be some init function to parse the version and commit out of the json set in that variable at compile time.

then cmd/core would just import "github.com/cogentcore/core" or "cogentcore.org/core" and then add a `-v --version` flag to the command line interface to print the version. And printing the version in the `--help` menu by default never hurts either. And it will still work without setting the version, just default to `unknown` version if the variable was not set on compilation.

...

I may submit this as a PR - if it would help. I thought of this for another project for which the versioning is more critical a concern ; so i'll be implementing it there already.

kkoreilly:
Use `debug.ReadBuildInfo` for displaying the version in the `cli` usage string, such as for the core tool. Also use `debug.ReadBuildInfo` as a backup for `system.AppVersion` and `system.CoreVersion` if those versions weren't already set by the core tool.

Also fixes a cursor build error on web caused by #1372.

For #1370. This is not a full fix as discussed there.

### DIFF
--- a/cli/usage.go
+++ b/cli/usage.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"runtime/debug"
 	"slices"
 	"strconv"
 	"strings"
@@ -47,6 +48,19 @@ func usage[T any](opts *Options, cfg T, cmd string, cmds ...*Cmd[T]) string {
 			fmt.Println(logx.CmdColor(cmdName()+" help") + logx.ErrorColor(fmt.Sprintf(" failed: command %q not found", cmd)))
 			os.Exit(1)
 		}
+	}
+
+	if bi, ok := debug.ReadBuildInfo(); ok {
+		revision, time := "dev", "unknown"
+		for _, set := range bi.Settings {
+			if set.Key == "vcs.revision" {
+				revision = set.Value
+			}
+			if set.Key == "vcs.time" {
+				time = set.Value
+			}
+		}
+		b.WriteString(logx.TitleColor("Version: ") + fmt.Sprintf("%s (%s)\n\n", revision, time))
 	}
 
 	fs := &fields{}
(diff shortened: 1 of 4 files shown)


rcoreilly (approved):
ya gut

Issue #1012, opened 2024-07-14
Text in a frame that centers its content, contained within a grid, is not rendered


kkoreilly:
### Describe the bug

See TestRenderGridCenteredFrame.

### How to reproduce

See TestRenderGridCenteredFrame.

### Platform

macOS

Issue #1098, opened 2024-07-31
Form is not fully robust to updating documentation and default tag in rare cases


kkoreilly:
### Describe the bug

As discussed in #1096, there are ways to break the `Form` updating logic for `default` struct tags and field documentation.

### How to reproduce

If you set the struct of a Form to a struct that is otherwise the same but has different `default` struct tags or different documentation, it will not update accordingly, since the `default` tags and documentation are only computed in an init function, which will not be called again if the name, type, and index of the field do not change. That is an exceedingly rare possibility and not worth fixing now, but we should fix it eventually; all of the doc, readOnlyTag, def, hasDef variables etc need to be recomputed every time in every relevant nested closures instead of just being computed once in an init function.

### Platform

macOS

kkoreilly:
#1389 did not fix this rare issue as discussed here: https://github.com/cogentcore/core/pull/1389#discussion_r1896472635.

PR #1389, opened 2024-12-21
Fix for #1387 -- layout logic says that bounding box should be based on Alloc, not Actual; also fix form stale closure issue using field path

### LINKED ISSUE #1387: form layout is too small from at least multiple spinner elements in add-fields. Present already in 0.3.0 initial release.
### Describe the bug

axon.LayerParams and PathParms exhibit the problem. 

when I comment-out the shouldDisplay logic, the problem goes away.

Also, shouldDisplay inline fields are causing crashing due to stale captured values in the maker updater.

### How to reproduce

run current axon, edit LayerParams or PathParams

working on making a reproducible test case

### Example code

_No response_

### Relevant output

_No response_

### Platform

macOS

### LINKED ISSUE #1390: form shouldDisplay logic causes crash due to stale captured variables -- updated demo to reveal
### Describe the bug

form contains code like this inside field closures:

```Go
					w.Styler(func(s *styles.Style) {
						f := fm.structFields[i]
```

but the value of `i` depends on what was actually  shown previously, so it can crash with out-of-range errors here. this happens in axon, and I updated the order of fields in demo to show it.

we need a canonical path name, not an index, and to find the field data by that path name.

### How to reproduce

open demo, click things in the first inline guy.

### Example code

_No response_

### Relevant output

_No response_

### Platform

macOS

rcoreilly:
Fixes #1387, fixes #1390


core/layout.go
```diff
@@ -295,13 +295,13 @@ func (ls *geomState) contentRangeDim(d math32.Dims) (cmin, cmax float32) {
 // totalRect returns Pos.Total -- Size.Actual.Total
 // as an image.Rectangle, e.g., for bounding box
 func (ls *geomState) totalRect() image.Rectangle {
-	return math32.RectFromPosSizeMax(ls.Pos.Total, ls.Size.Actual.Total)
```
kkoreilly, line 297: We had the previous behavior for a reason; I think it might have been related to border, shadow, scrollbar, or parent border radius rendering or something. See https://github.com/cogentcore/core/commit/c3ca7fa41d79826f21612532fa7524432e35bee8; you must test very thoroughly.

rcoreilly: The only tests that differ in my local run were 2 scrollbar things that seem like they were incorrect before.

kkoreilly: This was the commit I was thinking of where we previously switched to using actual in another context to fix #1012: 316494b28b932079d29ccaba87406a1a2ef54323

rcoreilly: I just pushed a key fix to remove the spacing change in my form test. You can verify that the test we wrote for that issue is _not_ affected by this change. Furthermore if you look at the layout logic, it is essential that the bounding box be given by the alloc, because that is what is used for arranging items within a frame. Most of the time we weren't noticing this issue because most frames that do significant layout are also grow, which eliminates this differnece between actual and alloc, so this was showing up in form because we turned grow off.

rcoreilly:
ps will rename Actual -> Need in a later PR to make it clearer what it "actually" is.

rcoreilly:
ok now actually fully ready  :)

core/form.go
```diff
@@ -109,10 +110,11 @@ func (fm *Form) getStructFields() {
 						if field.Tag.Get("edit") == "-" && sfield.Tag.Get("edit") == "" {
 							sfield.Tag += ` edit:"-"`
 						}
-						fields = append(fields, &structField{path: field.Name + " • " + sfield.Name, field: sfield, value: value, parent: parent})
+						path := field.Name + " • " + sfield.Name
+						fields.Add(path, &structField{path: path, field: sfield, value: value, parent: parent})
```
kkoreilly, line 114: We are using the user-friendly bullet text for the map key?

core/form.go
```diff
@@ -167,7 +171,7 @@ func (fm *Form) Init() {
 			// as the ground truth everywhere. The index could otherwise become invalid,
 			// such as when a ShouldDisplayer condition is newly satisfied
 			// (see https://github.com/cogentcore/core/issues/1096).
```
kkoreilly: This documentation is out-of-date.

core/form.go
```diff
@@ -190,7 +194,7 @@ func (fm *Form) Init() {
 					}
```
kkoreilly, line 193: You are still not recomputing doc, readOnlyTag, def, hasDef, etc every time. This does **not** address #1098; that really is for rare cases, which are different from those this PR addresses.

kkoreilly (changes requested):

kkoreilly (approved):

PR #1402, opened 2024-12-25
fileinfo: only get mimeinfo and generated from regular files

rcoreilly:
This should fix the CD of pages / docs, and generally be more robust.

*Edit from @kkoreilly: this supersedes and thus closes #1395, which was useful for debugging but did not contain a fix*

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -31,6 +31,9 @@ jobs:
     - name: Test
       run: go test -v $(go list ./... | grep -v gpu | grep -v xyz | grep -v system) -coverprofile cover.out -timeout 30s
 
+    - name: Build Docs test
+      run: core build web -dir docs -o static
+
     - name: Update coverage report
       uses: ncruces/go-coverage-report@v0
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -28,9 +28,15 @@ jobs:
...
-    - name: Test
-      run: go test -v $(go list ./... | grep -v gpu | grep -v xyz | grep -v system) -coverprofile cover.out -timeout 30s
+#    - name: Test
+#      run: go test -v $(go list ./... | grep -v gpu | grep -v xyz | grep -v system) -coverprofile cover.out -timeout 30s
 
+    - name: Install Core
+      run: go install ./cmd/core
+
+    - name: Build Docs
+      run: core build web -dir docs -o static -vanity-url cogentcore.org/core -github-vanity-repository cogentcore/core
```
kkoreilly: Please add this logic back for reliable testing purposes. You do **not** need to install the core tool again (it already is above). Instead of `Build Docs`, say `Build Docs (to verify it works on PRs)`, and you do not need all the vanity stuff.

kkoreilly (changes requested):

kkoreilly (approved):

PR #1432, opened 2025-01-15
refactor: removes useless `else` to get more idiomatic GO code

chavacava:
This PR removes useless `else` statements and replaces code blocks to get more idiomatic Go code and less deep nesting.

Refs: https://youtu.be/5DVV36uqQ4E?t=660

### DIFF
--- a/base/labels/friendly.go
+++ b/base/labels/friendly.go
@@ -42,9 +42,8 @@ func FriendlyTypeName(typ reflect.Type) string {
 			return "List of " + bnm
 		} else if strings.Contains(bnm, "Function of") {
 			return strings.ReplaceAll(bnm, "Function of", "Functions of") + "s"
-		} else {
-			return bnm + "s"
 		}
+		return bnm + "s"
 	case reflect.Func:
 		str := "Function"
 		ni := nptyp.NumIn()
--- a/base/reflectx/pointers.go
+++ b/base/reflectx/pointers.go
@@ -64,11 +64,11 @@ func OnePointerValue(v reflect.Value) reflect.Value {
 		pv := reflect.New(v.Type())
 		pv.Elem().Set(v)
 		return pv
-	} else {
-		for v.Elem().Kind() == reflect.Pointer {
-			v = v.Elem()
-		}
 	}
+	for v.Elem().Kind() == reflect.Pointer {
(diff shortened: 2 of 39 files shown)


AnyCPU:
some notes:

1) actually there are a lot of good code style guides from Google, Uber etc, but yes - some `else` is redundant;
2) this pr is huge, maybe it is ok to decouple something;
3) some if-chains could be written as `switch`es, for instance:

Origin:
```golang
if strings.HasSuffix(bnm, "s") {
	return "List of " + bnm
} else if strings.Contains(bnm, "Function of") {
	return strings.ReplaceAll(bnm, "Function of", "Functions of") + "s"
} else {
	return bnm + "s"
}
```

Result:
```golang
switch {
    case strings.HasSuffix(bnm, "s"):
        return "List of " + bnm

    case strings.Contains(bnm, "Function of"):
        return strings.ReplaceAll(bnm, "Function of", "Functions of") + "s"

    default:
        return bnm + "s"
}
```

or maybe someone would like to introduce a new string var and use single `return`.

chavacava:
Hi @AnyCPU thanks for commenting on the PR

>     1. actually there are a lot of good code style guides from Google, Uber etc, but yes - some `else` is redundant;

Indeed. Anyway using `else`  is more the exception than the rule in GO codebases. (I'm not _in favor of_ or _against to_ ` else`, it's just the way idiomatic GO is)  
 
>     2. this pr is huge, maybe it is ok to decouple something;

Yes is a big PR. I don't know which is the level of confidence the PR checks provide. If the testing and checks infrastructure is not mature enough to merge without stress then yes, the PR need to be decoupled.

>     3. some if-chains could be written as `switch`es, for instance:

Yes, it was my next step after cleaning these if-else chains (my bet is that removing these chains will ease identifying where to refactor by introducing a `switch`)

kkoreilly:
@AnyCPU @chavacava Thank you for the feedback. I think that the PR is fine in terms of size, as the changes are clear and the CI would catch any major problems (#1347 will make the tests even more robust). It would definitely be nice to use switch statements in some of these contexts, but that is fine as a separate PR, so I will merge this now.

kkoreilly (approved):
Thanks for this contribution!

PR #1434, opened 2025-01-16
gpu/bind_group: added safe finalizers for CGO allocated objects

### LINKED ISSUE #1430: memory overflow or crash when continuously scaling windows
### Describe the bug

os: windows 11
go version: 1.23

memory overflow or crash when continuously scaling windows.

before: 

after:

up to 1.6GB of memory usage, but I only ran a helloworld. and the initialization of the program actually occupies 70M of memory.

### How to reproduce

Just need to continuously zoom in and out of the window.

### Example code

```go
package main

import (
	"cogentcore.org/core/core"
)

func main() {
	b := core.NewBody()
	core.NewButton(b).SetText("Hello, World!")
	b.RunMainWindow()
}
```

### Relevant output

_No response_

### Platform

Windows

AnyCPU:
This is actual for the Bind Group.

Fixes https://github.com/cogentcore/core/issues/1430

### DIFF
--- a/gpu/cpipeline.go
+++ b/gpu/cpipeline.go
@@ -134,8 +134,9 @@ func (pl *ComputePipeline) BindAllGroups(ce *wgpu.ComputePassEncoder) {
 func (pl *ComputePipeline) BindGroup(ce *wgpu.ComputePassEncoder, group int) {
 	vs := pl.Vars()
 	vg := vs.Groups[group]
-	bg, dynOffs, err := vg.bindGroup(vs)
+	bg, release, dynOffs, err := vg.bindGroup(vs)
 	if err == nil {
+		defer release()
 		ce.SetBindGroup(uint32(vg.Group), bg, dynOffs)
 	}
 }
--- a/gpu/gpipeline.go
+++ b/gpu/gpipeline.go
@@ -27,6 +27,8 @@ type GraphicsPipeline struct {
 
 	// renderPipeline is the configured, instantiated wgpu pipeline
 	renderPipeline *wgpu.RenderPipeline
+
+	finalizers []func() error
 }
 
 // NewGraphicsPipeline returns a new GraphicsPipeline.
@@ -56,10 +58,17 @@ func (pl *GraphicsPipeline) BindAllGroups(rp *wgpu.RenderPassEncoder) {
 func (pl *GraphicsPipeline) BindGroup(rp *wgpu.RenderPassEncoder, group int) {
 	vs := pl.Vars()
 	vg := vs.Groups[group]
-	bg, dynOffs, err := vg.bindGroup(vs)
+	bg, release, dynOffs, err := vg.bindGroup(vs)
 	if err == nil {
+		pl.finalizers = append(pl.finalizers, release)
 		rp.SetBindGroup(uint32(vg.Group), bg, dynOffs)
 	}
+}
+
+func (pl *GraphicsPipeline) DrainFinalizers() []func() error {
+	finalizers := pl.finalizers
+	pl.finalizers = nil
+	return finalizers
 }
 
 // BindPipeline binds this pipeline as the one to use for next commands in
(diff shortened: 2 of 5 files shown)


rcoreilly:
@AnyCPU thanks for the fix here!

did you try just doing:

```Go
	if vg.currentBindGroup != nil {
		vg.currentBindGroup.Release()
	}
```

in the `bindGroup` method? I just did that and it works fine on mac..

AnyCPU:
@rcoreilly yes, i did.

the simplest hello world example works well in that case, but the `demo` code crashes when you try to switch between tabs.

in case of the `demo`, you will see a stack trace from `C` code or `Rust`s bindings.

@rcoreilly that's why i added some comments (not many, but).

AnyCPU:
@rcoreilly and it should be noted, i'm pretty new with this code base, so it is possible there is better solution for this issue.

in the beginning of my journey i thought who was responsible to gather and execute `finalizers`.

i thought about the `RenderPassEncoder`, but it looks like it is not its responsibility.

rcoreilly:
@AnyCPU I just pushed #1435 which provides a more general solution to the leak. The system `SubmitRender` call is when everything should be done, so it is safe to release. This issue is more general than gpudrawer so this deals with it at the base gpu system level. Again thanks so much for finding the leak and providing your solution which really helped!

rcoreilly:
@AnyCPU how did you find this leak in the first place? I'm still seeing leaks with both your code and my version, for resizing. The Go pprof memory profiler can't see the cgo code, and I can't get the xcode instruments program to work with my go built file (needs signing and some kind of flag).

rcoreilly:
@AnyCPU nevermind -- I just went ahead and turned off my system integrity protections and now it works :) 
We are still leaking surface textures!

AnyCPU:
@rcoreilly 
> how did you find this leak in the first place? 

i've some experience in Windows GDI+, that time i learned a one rule: if you get some resource you must put it back or clean up. i have not so much experience in low level programming on macOS, so, i tried to use what i already know Go's profiler and easy-to-use command line utility - `leaks`. the `leaks` shows a lot of calls to two specific C functions. because i was pretty lazy to build CGO code in debug mode, i just investigated code around where related C functions were used. I guess it was much faster than rebuilding related packages in debug mode.

> We are still leaking surface textures!

i guess it is true, but now it happens really slowly. And a lot of apps can be run just fine.

**what can i do?**
besides the bind groups i also investigated the `system.drawer` impl and the `slicesx` package.

the drawer's code looks pretty suspicious to me: it looks like some underlying slices never shrink and some items never GC collected.

the `gpudraw.images` additionally has some troubles like in the `images.start`: iteration over a map of the `all` images.
the trick is that Go's map is unordered. Every time when you write `... := range map` you will get items in random order.
It might not work with code where slices are used as stack (last input first output).

there are two options available: preserve an order of item where it is required or use Go's maps as sets and rewrite some logic.

also the `slicesx` package might be reviewed: `GrowTo` and `SetN` functions, they look not so good. Maybe some Go's tests and benchmarks should be written.

everything above is just good options to start to improve.

i have no good examples or use cases to demo troubles above at the moment.

some options to improve memory leak investigation:
- Go's profiler
- debug mode
- https://github.com/iovisor/bcc/blob/master/tools/memleak.py
- https://github.com/jemalloc/jemalloc/wiki/Use-Case:-Leak-Checking

in general the project works much better today than yesterday.

kkoreilly:
@AnyCPU Thank you again for your contribution, which was very helpful for figuring out how to fix this. If the main resize leaking issue is fixed for you by #1435, please feel free to close this PR.

The main cause of remaining leaking when zooming is the font opening as discussed here: https://github.com/cogentcore/core/issues/1430#issuecomment-2599009831. However, if you do find any other leaks, such as those to do with `system` or `slicesx` as you discussed above, then it would definitely be good to fix them. I will let @rcoreilly write a more detailed response to the specifics of functions such as `slicesx.Grow`.

PR #1435, opened 2025-01-16
record and release old wgpu BindGroup items, which were otherwise not being released.

### LINKED ISSUE #1430: memory overflow or crash when continuously scaling windows

rcoreilly:
Thanks to @AnyCPU for identifying leak and potential fix in #1434, and @StinkyPeach for reporting the issue.

Relevant to #1430

note that with screen zoom that there is a remaining issue associated with loading fonts for each different DPI size, which is not fixed with this PR, and will not be addressed until we do the vello and go-text refactor.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,10 +12,9 @@ require (
 	github.com/bramvdbogaerde/go-scp v1.4.0
 	github.com/chewxy/math32 v1.10.1
 	github.com/cogentcore/reisen v0.0.0-20240814194831-4d884b6e7666
-	github.com/cogentcore/webgpu v0.0.0-20241212004832-ad7475f3b4dd
+	github.com/cogentcore/webgpu v0.0.0-20250117020559-3809548f4891
 	github.com/cogentcore/yaegi v0.0.0-20240724064145-e32a03faad56
 	github.com/coreos/go-oidc/v3 v3.10.0
-	github.com/ergochat/readline v0.1.2
 	github.com/ericchiang/css v1.3.0
 	github.com/faiface/beep v1.1.0
 	github.com/fsnotify/fsnotify v1.7.0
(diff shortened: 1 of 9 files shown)


rcoreilly:
I have now verified using the xcode instruments leak detector (had to turn system integrity protection off to run on go binaries) that this PR does fix the resizing memory leaks. 

However, the special `libwgpu_native.a` library that I built using a PR to access a more recent wgpu version, which is NOT released in our `webgpu` package, does still leak surface textures. This library runs compute jobs significantly faster, but is not anything official, so hopefully they will come out with updated wgpu_native packages sometime soon, that actually work.

Even with the "good" released version of the wgpu_native library, there _is_ an initial bump in memory usage at the point of first resizing, but it stabilizes from that point onward.

kkoreilly (changes requested):
LGTM overall, and the memory leak does seem to be gone for me when resizing, but there is still an unequivocal persistent memory leak when zooming the window in and out, which was the original report in #1430.

kkoreilly:
Note that the same zooming memory leak appears to occur in #1434 as well. The fix could be a separate PR, but this PR then would not be a fix for #1430.

rcoreilly:
it is the same situation as the resize:

image

an initial bump and then transients that re-stabilize. this plot from xcode leaks looks identical to resize.

kkoreilly (approved):

kkoreilly:
See https://github.com/cogentcore/core/issues/1430#issuecomment-2599009831 for information about the remaining leaking when zooming.

PR #1440, opened 2025-01-18
renderwindow resized is now much more parsimonious if everything is the same

### LINKED ISSUE #1438: Moving a window ends up calling resized() with same size: need to do less when that happens
### Describe the bug

see renderwindow.go: 284 case where curRg == rg -- this causes any open dialogs with textfield to lose their edits, for example.

is only rarely actually necessary to do something here: figure out the right conditions and only do it then!

also cogent code is doing too much here: tree seems to be redoing git repo file loading which is slow, for example -- trace it.

### How to reproduce

move windows

### Example code

```go

```

### Relevant output

```shell

```

### Platform

macOS

rcoreilly:
Fixes #1438

### DIFF
--- a/core/renderwindow.go
+++ b/core/renderwindow.go
@@ -284,17 +284,22 @@ func (w *renderWindow) resized() {
 		if DebugSettings.WindowEventTrace {
 			fmt.Printf("Win: %v skipped same-size Resized: %v\n", w.name, curRg)
 		}
-		rc.logicalDPI = w.logicalDPI()
-		w.mains.resize(rg) // no-op if everyone below is good
-		// still need to apply style even if size is same
-		for _, kv := range w.mains.stack.Order {
-			st := kv.Value
-			sc := st.Scene
-			if st.FullWindow && sc.SceneGeom.Size != rg.Size { // double-check: can be off in fullscreen init
-				st.Sprites.reset()
-				sc.resize(rg)
+		newDPI := false
+		if rc.logicalDPI != w.logicalDPI() {
+			rc.logicalDPI = w.logicalDPI()
+			newDPI = true
+		}
+		if w.mains.resize(rg) || newDPI {
+			// still need to apply style even if size is same
+			for _, kv := range w.mains.stack.Order {
+				st := kv.Value
+				sc := st.Scene
+				// if st.FullWindow && sc.SceneGeom.Size != rg.Size { // double-check: can be off in fullscreen init
+				// 	st.Sprites.reset()
+				// 	sc.resize(rg)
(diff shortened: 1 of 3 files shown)


core/renderwindow.go
```diff
@@ -284,17 +284,22 @@ func (w *renderWindow) resized() {
...
-		// still need to apply style even if size is same
-		for _, kv := range w.mains.stack.Order {
-			st := kv.Value
-			sc := st.Scene
-			if st.FullWindow && sc.SceneGeom.Size != rg.Size { // double-check: can be off in fullscreen init
-				st.Sprites.reset()
-				sc.resize(rg)
+		newDPI := false
+		if rc.logicalDPI != w.logicalDPI() {
+			rc.logicalDPI = w.logicalDPI()
```
kkoreilly: Is `w.logicalDPI()` sufficiently nontrivial to justify a local variable here?

core/renderwindow.go
```diff
@@ -284,17 +284,22 @@ func (w *renderWindow) resized() {
...
+		}
+		if w.mains.resize(rg) || newDPI {
+			// still need to apply style even if size is same
+			for _, kv := range w.mains.stack.Order {
+				st := kv.Value
+				sc := st.Scene
+				// if st.FullWindow && sc.SceneGeom.Size != rg.Size { // double-check: can be off in fullscreen init
+				// 	st.Sprites.reset()
+				// 	sc.resize(rg)
+				// }
```
kkoreilly: What is happening here?

rcoreilly:
This eliminates the long lags in cogent code from resizing -- there were multiple resized events being generated at the same size and it was redoing everything multiple times. also same with general resize resource blips.

kkoreilly (changes requested):

core/renderwindow.go
rcoreilly: right thanks -- need to nuke that now that it is working.

rcoreilly: was there something different about fullscreen on windows vs. mac though? do I need to test it there?

kkoreilly: It does seem to work for me on mac with the initial fullscreen state. I don't think there is anything different on Windows. Can you explain why this PR removes the need for that code? It was definitely needed before.

rcoreilly: ok tested there just for good measure -- looks good.

core/renderwindow.go
rcoreilly: ?

kkoreilly (approved):

PR #1321, opened 2024-11-18
Add Cogent Content, a new system for managing and displaying static web content

kkoreilly:
This is a successor to `pages`, based on the design of Hugo and MediaWiki. This will allow us to improve our documentation.

PR #1444, opened 2025-01-21
valuer.go toValue: do not use a text representation for nil pointers

rcoreilly:
because List and Table widgets will make those for empty rows, but when they later are used, the widget remains a text widget which is unacceptable behavior. fixed a crash on TreeButton -- any other issues can be fixed in a similar manner.


core/valuer.go
```diff
@@ -85,9 +85,9 @@ func toValue(value any, tags reflect.StructTag) Value {
 		return NewText()
 	}
 	uv := reflectx.Underlying(rv)
-	if reflectx.IsNil(uv) {
-		return NewText()
-	}
+	// if reflectx.IsNil(uv) { // no: handle null properly in relevant widgets; essential for list / table
```
AnyCPU: @rcoreilly usually i do not recommend to leave such comments, on my opinion it adds more cognitive complexity. if it is important it is ok to keep related info in doc comments before function signature.

just my 2 cents

AnyCPU: code itself could be removed, if someone is interested in then it is always possible to take a look at git history.

rcoreilly: Yes good point -- this was here for @kkoreilly's benefit in reviewing but I would probably forget to remove the comment and now that I've made this PR with all the discussion, clearly this is the proper location for such comments!

kkoreilly (commented):
@rcoreilly I see now that you merged this PR into #1321 already, so I will close this PR.

However, when testing this I noticed two things:

1. The parts `None` button will still open a dialog, which it shouldn't.
2. The properties map now correctly updates when switching from nil to an extant map, but it does *not* update when switching between two different extant maps (run the inspector in the docs and select a frame and then a text element to see this effect).

Issue #1051, opened 2024-07-25
Support character-specific text selection


kkoreilly:
### Describe the feature

Similar to how it currently works in text fields and text editors.

kkoreilly:
We should also figure out text selection in read only text fields as part of this.

PR #1448, opened 2025-01-24
logic to only scroll when focused, so texteditors don't keep grabbing everything when scrolling

rcoreilly:
* added abilities.RemoteScrollable that determines whether a Frame can be scrolled without being focused. It is on by default for Frame, and off for texteditor.Editor.

* todo: need to allow slide scrolling to propagate upward and not get trapped in the texteditor.


rcoreilly:
`RemoteScrollable` is admittedly a bit of a strange name -- also considered `NoFocusScrollable` or `ScrollableNoFocus`. 

maybe `ScrollableOnlyFocused` or `ScrollableWhenFocused` ?   last one sounds best probably.

AnyCPU:
@rcoreilly i know little about related logic, but a `ScrollableUnfocused` option might be considered, might not?

kkoreilly (changes requested):
@AnyCPU I agree that something like `ScrollableUnfocused` or `ScrollableWithoutFocus` would be a better name. Thanks for the suggestion.

However, I question the overall approach of this PR, @rcoreilly. You are defining a general ability that is then eclectically applied in some subset of widgets to an arbitrary and varying set of events (Scroll, SlideMove, MouseDown, MouseUp, DoubleClick, TripleClick, etc). You are broadly construing this ability to do what you want, instead of implementing a consistent definition.

Given that the true purpose of this PR is to prevent handling of mouse and slide events on unfocused text editors, I would suggest just implementing that logic directly in a much more concise and elegant way. The handling of unfocused scroll events in particular is not actually a major problem, and is in fact intended behavior in many cases.

rcoreilly:
@kkoreilly an important issue here is that the texteditor doesn't handle its own scroll events -- those come from the general frame logic that it inherits. And some of the logic needs to be in the general events handler. If you can figure out how to get everything to work without the overall abilities flag, which makes semantic sense to me as a coherent ability btw, then go ahead. You should also actually try the behavior here on a mobile device and see what a huge difference it makes in usability.

kkoreilly:
@rcoreilly Please see the last paragraph of my comment:

> Given that the true purpose of this PR is to prevent handling of mouse and slide events on unfocused text editors, I would suggest just implementing that logic directly in a much more concise and elegant way. The handling of unfocused scroll events in particular is not actually a major problem, and is in fact intended behavior in many cases.

The issue on mobile is *not* about scrolling, it is about mouse and slide events. Scrolling events *should* be handled when unfocused (such as in Cogent Code), and so I am suggesting that you just require focus for the mouse and slide events directly in the text editor logic, without any of this abilities stuff.

core/events.go
```diff
@@ -322,10 +322,13 @@ func (em *Events) handlePosEvent(e events.Event) {
 			if press == nil && (wb.Styles.Abilities.IsPressable() || sc.renderBBoxes) {
 				press = w
 			}
-			if dragPress == nil && wb.Styles.Abilities.Is(abilities.Draggable) {
+			if dragPress == nil && wb.AbilityIs(abilities.Draggable) {
 				dragPress = w
 			}
-			if slidePress == nil && wb.Styles.Abilities.Is(abilities.Slideable) {
+			if slidePress == nil && wb.AbilityIs(abilities.Slideable) {
+				if !wb.AbilityIs(abilities.ScrollableUnfocused) && !wb.Styles.Is(states.Focused) {
```
kkoreilly: Use `wb.StateIs`

core/frame.go
```diff
@@ -142,15 +143,22 @@ func (fr *Frame) Init() {
 		fr.focusOnName(e)
 	})
 	fr.On(events.Scroll, func(e events.Event) {
-		fr.scrollDelta(e)
+		if fr.AbilityIs(abilities.ScrollableUnfocused) || fr.Styles.Is(states.Focused) {
```
kkoreilly: I thought that the handling of this check was going to be centralized in the event manager?

core/frame.go
```diff
@@ -142,15 +143,22 @@ func (fr *Frame) Init() {
...
+			fr.scrollDelta(e)
+		}
 	})
 	// We treat slide events on frames as scroll events.
 	fr.On(events.SlideMove, func(e events.Event) {
 		// We must negate the delta for "natural" scrolling behavior.
-		del := math32.FromPoint(e.PrevDelta()).MulScalar(-0.034)
-		fr.scrollDelta(events.NewScroll(e.WindowPos(), del, e.Modifiers()))
+		if fr.AbilityIs(abilities.ScrollableUnfocused) || fr.Styles.Is(states.Focused) {
```
kkoreilly: Isn't this check unnecessary?

core/frame.go
```diff
@@ -142,15 +143,22 @@ func (fr *Frame) Init() {
...
+			del := math32.FromPoint(e.PrevDelta()).MulScalar(-0.034)
+			fr.scrollDelta(events.NewScroll(e.WindowPos(), del, e.Modifiers()))
+		}
 	})
 	fr.On(events.SlideStop, func(e events.Event) {
+		if !(fr.AbilityIs(abilities.ScrollableUnfocused) || fr.Styles.Is(states.Focused)) {
```
kkoreilly: Unnecessary.

styles/abilities/abilities.go
```diff
@@ -73,6 +73,10 @@ const (
 
 	// LongHoverable means it can be LongHovered
 	LongHoverable
+
+	// ScrollableUnfocused means it can be Scrolled and Slided without focus.
+	// This is true by default except for text editors and text fields.
```
kkoreilly, line 78: What about making this off by default for plots? Those are another widget that always intercept my scrolling.

texteditor/events.go
```diff
@@ -539,30 +539,17 @@ func (ed *Editor) OpenLinkAt(pos lexer.Pos) (*paint.TextLink, bool) {
 
 // handleMouse handles mouse events
 func (ed *Editor) handleMouse() {
-	ed.On(events.MouseDown, func(e events.Event) { // note: usual is Click..
-		if !ed.StateIs(states.Focused) {
-			ed.SetFocus()
-		}
+	ed.OnClick(func(e events.Event) {
+		ed.SetFocus()
```
kkoreilly, line 543: Why focus for read-only text editors but not read-only text fields?

texteditor/events.go
```diff
@@ -595,11 +582,24 @@ func (ed *Editor) handleMouse() {
 		}
 		ed.NeedsRender()
 	})
-	ed.On(events.SlideMove, func(e events.Event) {
+	ed.On(events.SlideStart, func(e events.Event) { // note: usual is Click..
```
kkoreilly: Is the comment really necessary?

kkoreilly (changes requested):
Overall this seems like a good solution.

core/frame.go
rcoreilly: There is no central management of events.Scroll per se and Frame is the only scroller..

styles/abilities/abilities.go
rcoreilly: Plots are in lab, so we can address that there -- we would need to add a focus state -- not unreasonable because we could also add keyboard shortcuts for plotting probably.

texteditor/events.go
rcoreilly: we'll deal with this later in context of broader text selection revamp, after new text rendering infra.

rcoreilly: #1051

core/frame.go
```diff
@@ -142,7 +143,9 @@ func (fr *Frame) Init() {
 		fr.focusOnName(e)
 	})
 	fr.On(events.Scroll, func(e events.Event) {
-		fr.scrollDelta(e)
+		if fr.AbilityIs(abilities.ScrollableUnfocused) || fr.Styles.Is(states.Focused) {
```
kkoreilly: `fr.StateIs`.

kkoreilly (changes requested):

kkoreilly (approved):

Issue #1458, opened 2025-01-26
Windows: an app freezes forever (not responding) if ContentScale/DevicePixelRatio is equal or great than 1.5


AnyCPU:
### Describe the bug

Windows: an app freezes forever (not responding) if ContentScale/DevicePixelRatio is equal or great than 1.5

### How to reproduce

1. configure the Cogent Core for Windows 64 bit;
2. try to apply custom scaling options in `Settings` on Windows, for instance, use: `150%` factor;
3. or connect to Windows machine from macOS via a `Windows App` (Microsoft Remote Desktop) with an option `Optimize for Retina displays` turned on;
4. try to run the `hello world`;
5. open either `Settings` or `Inspector`.

my environment:
- Go: go1.23.5 windows/amd64
- Zig: 0.14.0-dev.2851+b074fb7dd
- Windows 10 (22H2) 64 bit

useful commands:
- go install cogentcore.org/core/cmd/core@main
- core setup
- go env -w CGO_ENABLED=1 (by default this option is turned off on WIndows 10)
- go env -w CC="C:\zig0.14\zig cc"
- go env -w CXX="C:\zig0.14\zig c++"
- go env -w CGO_LDFLAGS="-O2 -g -lunwind" (this is important, on Windows 10 64 bit we must add -lunwind)

### Example code

```go
package main

import (
	"cogentcore.org/core/core"
	"cogentcore.org/core/events"
)

func main() {
	b := core.NewBody()
	btn := core.NewButton(b).SetText("Message")
	btn.OnClick(func(e events.Event) {
		core.MessageDialog(btn, "Hello, World!", "Message")
	})
	b.RunMainWindow()
}
```

### Relevant output

```shell
An app title + (Not Responding)
```

### Platform

Windows

#### 20250129
Why it happens?
It happens when you connect to remote Windows machine from macOS via remote desktop client and then Windows OS sees Retina display. And then the `Core` gets into deadlock because of following functions:
- func (w *Window) ConstrainFrame(topOnly bool) styles.Sides[int]
- func (w *Window) SetGeometry(fullscreen bool, pos, size image.Point, screen *system.Screen)
- func (a *App) RunOnMain(f func())
- func (a *App) MainLoop() for desktop

this pr partially fixes an issue above (https://github.com/cogentcore/core/pull/1462), but new windows are big as twice as needed.

[...]

PR #1462, opened 2025-01-29
window/update_geometry: fixed assigning screen coords to wm coords

### LINKED ISSUE #1461: Window.updateGeometry: coord sys mismatch
### Describe the bug

func: `func (w *Window) updateGeometry()`

the `updateGeometry` tries to assign a size to a window (WnSize) in different coord system.

### How to reproduce

Just take a look at source code.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

macOS

AnyCPU:
Fixes https://github.com/cogentcore/core/issues/1461

### DIFF
--- a/system/driver/desktop/window.go
+++ b/system/driver/desktop/window.go
@@ -484,7 +484,7 @@ func (w *Window) updateGeometry() {
 	w.updateMaximized()
 	var wsz image.Point
 	wsz.X, wsz.Y = w.Glw.GetSize()
-	w.WnSize = wsz
+	w.WnSize = sc.WindowSizeFromPixels(wsz)
 	var fbsz image.Point
 	fbsz.X, fbsz.Y = w.Glw.GetFramebufferSize()
 	w.PixelSize = fbsz


kkoreilly (changes requested):
Thank you for your contribution. According to the [glfw documentation](https://www.glfw.org/docs/3.3/window_guide.html#window_size), the GetSize function already returns the window size in screen/window manager coordinates, and so it does not make sense to treat that like pixels and convert it further. Please let me know if I am misunderstanding something.

AnyCPU:
@kkoreilly i guess i have lost in terminology.

there https://github.com/cogentcore/core/issues/1458 i added an update.
```
It happens when you connect to remote Windows machine from macOS 
via remote desktop client and then Windows OS sees Retina display. 
And then the Core gets into deadlock because of following functions:

func (w *Window) ConstrainFrame(topOnly bool) styles.Sides[int]
func (w *Window) SetGeometry(fullscreen bool, pos, size image.Point, screen *system.Screen)
func (a *App) RunOnMain(f func())
func (a *App) MainLoop() for desktop

this pr partially fixes an issue above (https://github.com/cogentcore/core/pull/1462), 
but new windows are big as twice as needed.

to reproduce this issue you also have to delete following directory 
C:\Users\<user_name>\AppData\Roaming\Cogent Core. 
An app might work if it could find some "appropriate" settings (1 of 100 chance, optimist).
```

it feels like this is not the best solution for the issue, but it works somehow at least :-)

i have some another changes that should be tested, but they look more weird.

AnyCPU:
@kkoreilly ok, i know what is wrong.

the `mon.GetVideoMode()` reports following:

on local macOS 
```
GetContentScale: cscx=2 adcs=2
pixel size pixel_size=(2880,1800) dpr=2 vm="&{Width:1440 Height:900 RedBits:8 GreenBits:8 BlueBits:8 RefreshRate:60}"
```

when connected to Windows from macOS
```
GetContentScale: cscx=2 adcs=2
pixel size pixel_size=(5760,3600) dpr=2 vm="&{Width:2880 Height:1800 RedBits:8 GreenBits:8 BlueBits:8 RefreshRate:32}"
```

i guess that is why a `Core` app gets into deadlock while i'm using RDP connection.

in order to run the app as expected i need to changes this line
https://github.com/cogentcore/core/blob/86d4cdbc1ac0bc3f4c00fee30f3af2aa9b74e3c5/system/driver/desktop/window.go#L352

and it should look like:
```go
go w.SetGeometry(false, pos, sz, sc)
```

then the app runs well.

AnyCPU:
@kkoreilly i have no idea how to make the `GetVideoMode` report correct values while using RDP.

should i close this pr and add a new one with following changes?

```go
go w.SetGeometry(false, pos, sz, sc)
```

AnyCPU:
not relevant

PR #1464, opened 2025-01-29
window/set_title_bar_is_dark: used correct values in win32 api call

### LINKED ISSUE #1463: Window.SetTitleBarIsDark: correct values for win32 api call
### Describe the bug

func: `func (w *Window) SetTitleBarIsDark(isDark bool)`

actually it is pretty strange bug.

i discovered it while i was working on https://github.com/cogentcore/core/issues/1458

i noticed messages like following in logs: `error: operation completed successfully`.

i found the `SetTitleBarIsDark` function and added some improvements.

changes have comments, please take a look at related pr.

### How to reproduce

this is hard.

in general i should connect from macOS to remote Windows machine via some remote desktop client and try to turn on/off Dark Theme.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

Windows

AnyCPU:
Fixes https://github.com/cogentcore/core/issues/1463

### DIFF
--- a/system/driver/desktop/theme_windows.go
+++ b/system/driver/desktop/theme_windows.go
@@ -105,16 +105,31 @@ func (w *Window) SetTitleBarIsDark(isDark bool) {
 	if !w.IsVisible() {
 		return
 	}
-	hwnd := w.Glw.GetWin32Window()
 
-	dwm := syscall.NewLazyDLL("dwmapi.dll")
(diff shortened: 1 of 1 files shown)


kkoreilly (approved):
Thank you for your contribution! I tested this using a UTM VM and confirmed that this works. (The previous version also worked for me, but presumably this version is more robust/compatible).

PR #1466, opened 2025-01-29
window: fixed boolean errors in if conditions

### LINKED ISSUE #1465: Window.ConstrainFrame: boolean error in if condition
### Describe the bug

func: `func (w *Window) ConstrainFrame(topOnly bool) styles.Sides[int]`

i think the `Window.ConstrainFrame` and `windowGeometrySaver.record` functions have boolean/logical errors in if conditions like:

```go
if TheApp.Platform() == system.Windows && w.Pos.X == -32000 || w.Pos.Y == -32000 {
	return w.FrameSize
}
```

but it should be like:

```go
if TheApp.Platform() == system.Windows && (w.Pos.X == -32000 || w.Pos.Y == -32000) {
	return w.FrameSize
}
```

it was found during my work on https://github.com/cogentcore/core/issues/1458

### How to reproduce

this is especially hard on Windows.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

Windows

AnyCPU:
affected functions: Window.ConstrainFrame and windowGeometrySaver.record.

Fixes https://github.com/cogentcore/core/issues/1465

### DIFF
--- a/core/windowgeometry.go
+++ b/core/windowgeometry.go
@@ -261,7 +261,6 @@ func (ws *windowGeometrySaver) record(win *renderWindow) {
 	}
 	sc := win.SystemWindow.Screen()
 	pos := win.SystemWindow.Position(sc)
-
 	ws.mu.Lock()
 	if ws.settingNoSave {
 		if DebugSettings.WindowGeometryTrace {
--- a/core/windowgeometry.go
+++ b/core/windowgeometry.go
@@ -257,7 +257,8 @@ func (ws *windowGeometrySaver) record(win *renderWindow) {
 	}
 	sc := win.SystemWindow.Screen()
 	pos := win.SystemWindow.Position(sc)
-	if TheApp.Platform() == system.Windows && pos.X == -32000 || pos.Y == -32000 { // windows badness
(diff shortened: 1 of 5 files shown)


core/windowgeometry.go
```diff
@@ -257,7 +257,8 @@ func (ws *windowGeometrySaver) record(win *renderWindow) {
 	}
 	sc := win.SystemWindow.Screen()
 	pos := win.SystemWindow.Position(sc)
-	if TheApp.Platform() == system.Windows && pos.X == -32000 || pos.Y == -32000 { // windows badness
+	// TODO: come up with better name for a -32000 constant
```
kkoreilly: Thank you for this contribution! If you want, it would be great if you could add a `system.WindowsMinimizedPosition` constant or similar.

AnyCPU: @kkoreilly i've updated pr and added some new changes are supposed to make related code more logical and clear.

AnyCPU: i can revert those changes and just use `system.WindowsMinimizedPosition` if any.

PR #1471, opened 2025-02-01
use scaled frame buffer when needed

### LINKED ISSUE #1470: desktop/window: use Retina-aware/scaled framebuffer when needed
### Describe the bug

Let's just use `GLFW_SCALE_FRAMEBUFFER` (added in GLFW 3.4) when needed.

It has the same effect as `GLFW_COCOA_RETINA_FRAMEBUFFER `.

So we can use the `GLFW_COCOA_RETINA_FRAMEBUFFER` for now.

Ref: https://www.glfw.org/docs/latest/window_guide.html

### How to reproduce

Change content scale.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

Windows

AnyCPU:
fixes https://github.com/cogentcore/core/issues/1470

### DIFF
--- a/system/driver/desktop/window.go
+++ b/system/driver/desktop/window.go
@@ -22,6 +22,12 @@ import (
 	"github.com/go-gl/glfw/v3.3/glfw"
 )
 
+// ScaleFramebuffer is a workaround for a GLFW_SCALE_FRAMEBUFFER window hint introduced in GLFW 3.4,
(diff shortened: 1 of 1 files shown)


kkoreilly:
@AnyCPU Thank you for your contribution! We will look at your PRs, test them on our Windows device, and get back to you soon.

PR #1475, opened 2025-02-04
icons: unordered set for used icons

### LINKED ISSUE #1474: use an unordered set for icons.Used
### Describe the bug

An `icons.Used` collection of icons contains a list of actually used icons in a Core app.
Right now it is implemented as a map of bool, but it can be represented as Go-style unordered set.
Also it saves some bytes at runtime, because a `struct{}{}` takes zero bytes. 

### How to reproduce

https://go.dev/ref/spec#Size_and_alignment_guarantees

### Example code

```go

```

### Relevant output

```shell

```

### Platform

macOS

AnyCPU:
fixes: https://github.com/cogentcore/core/issues/1474

### DIFF
--- a/core/icon.go
+++ b/core/icon.go
@@ -69,7 +69,7 @@ func (ic *Icon) readIcon() {
 	if errors.Log(err) != nil {
 		return
 	}
-	icons.Used[ic.Icon] = true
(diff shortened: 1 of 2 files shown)


kkoreilly (approved):
Thanks for your contribution!

PR #1527, opened 2025-05-19
Only include a subset of icons in yaegicore to reduce binary size

### LINKED ISSUE #1473: Reduce binary sizes of yaegicore apps
### Describe the feature

Apps that import yaegicore, such as the documentation website, end up including every symbol in a lot of packages (especially `icons`), in addition to the base size of yaegi itself. That substantially increases the binary size, impacting load times on web. We should use https://github.com/Zxilly/go-size-analyzer to figure out what packages are contributing the most and what we can cut. If we could figure out some way to include fewer icons in yaegicore, that would help a lot.

### Relevant code

```go

```

kkoreilly:
Instead of including all icons in `yaegicore`, we only include a subset (the icons we use in docs). This reduces binary size by around 2 MB on native and 3 MB on web. That is around 700 KB on compressed wasm, which is around 5% of the total web binary size.

Fixes #1473


rcoreilly (approved):
note: can extend this manual list by adding to Symbols in this package.

PR #1528, opened 2025-05-19
updates for Cogent Canvas

### LINKED ISSUE #1314: Text Field Icon Sizing is SUPER Awkward
### Describe the bug

When styling a text field you can set the font size very intuitively with a styler with either `s.Font.Size.Pt(12)` or `s.Font.Size = units.Pt(12)` however this does not change the size of the trailing icon where present (for example those added by `AddClearButton()` or `SetTypePassword()`)

Instead it appears that one must nest calls to `tree.AddChildInit` two deep to get to the actual icon (see example code)
In addition one must hunt for the appropriate child name of "trail-icon" which is a `*core.Button` and not a `*core.Icon` (somewhat contrary to the conclusion that widget part names are not rocket science reached in https://github.com/cogentcore/core/issues/840)

If there's a better way to do this that I've just overlooked in my search please do let me know.

### How to reproduce

Create a new text field widget
Set the font size and observe icons that are included in the field do not follow the size change

### Example code

```go
core.NewText(usernameContainer).SetText("Username")
	usernameField := core.NewTextField(usernameContainer).AddClearButton()
	tree.AddChildInit(usernameField, "trail-icon", func(trailIcon *core.Button) {
		tree.AddChildInit(trailIcon, "icon", func(icon *core.Icon) {
			icon.Styler(func(s *styles.Style) {
				s.Font.Size.Pt(14)
				s.Text.LineHeight.Pt(14)
			})
		})
	})
```

### Relevant output

_No response_

### Platform

Linux

### LINKED ISSUE #1406: Tab label font size
### Describe the bug

Updating the font size of the tab label does not work.

### How to reproduce

bug report

### Example code

_No response_

### Relevant output

_No response_

### Platform

Linux

rcoreilly:
Some fixes needed to svg and other repos to make Cogent Canvas fully work.

Fixes #1314, fixes #1406


colors/image.go
```diff
@@ -24,6 +27,20 @@ func ToUniform(img image.Image) color.RGBA {
...
+// an image color, to ensure that it doesn't modify all users of this image.
+func CloneUniform(img image.Image) image.Image {
+	if img == nil {
+		return nil
+	}
+	u, ok := img.(*image.Uniform)
+	if !ok {
+		return img
+	}
+	return image.NewUniform(AsRGBA(u.C))
```
kkoreilly, line 41: No need to call `AsRGBA` here, right? Also why not use `ToUniform` to make this a more general function for converting any image to a uniform one?

core/enumgen.go
```diff
@@ -178,16 +178,16 @@ func (i *MeterTypes) UnmarshalText(text []byte) error {
 	return enums.UnmarshalText(i, text, "MeterTypes")
 }
 
-var _renderWindowFlagsValues = []renderWindowFlags{0, 1, 2, 3, 4, 5}
```
kkoreilly, line 181: It would be helpful for debugging print statements to still have this...

core/events.go
```diff
@@ -581,6 +583,37 @@ func (em *Events) updateHovers(hov, prev []Widget, e events.Event, enter, leave
 	return hov
 }
 
+// updateSpriteHovers updates the hovered sprites based on current
+// sprites in bounding box.
+func (em *Events) updateSpriteHovers(hov, prev []*Sprite, e events.Event, enter, leave events.Types) []*Sprite {
+	for _, prv := range prev {
+		stillIn := false
```
kkoreilly: Use `slices.Contains`!

core/render_js.go
```diff
@@ -7,11 +7,37 @@
...
+	"cogentcore.org/core/math32"
+	"cogentcore.org/core/paint"
 	"cogentcore.org/core/paint/renderers/htmlcanvas"
 	"cogentcore.org/core/system/composer"
+	"golang.org/x/image/draw"
 )
 
+// grabRenderFrom grabs the rendered image from the given widget.
+// If it returns nil, then the image could not be fetched.
+func grabRenderFrom(w Widget) *image.RGBA {
```
kkoreilly, line 22: We should fetch the image from the canvas instead of manually rendering it.

core/render_notjs.go
```diff
@@ -16,6 +16,23 @@ import (
 	"golang.org/x/image/draw"
 )
 
+// grabRenderFrom grabs the rendered image from the given widget.
+// If it returns nil, then the image could not be fetched.
+func grabRenderFrom(w Widget) *image.RGBA {
+	wb := w.AsWidget()
+	scimg := wb.Scene.renderer.Image() // todo: need to make this real on JS
```
kkoreilly: Remove this todo and resolve it according to the comment in the js one.

core/rendersource.go
```diff
@@ -57,29 +60,28 @@ type scrimSource struct {
...
-		// note: may need to copy pixels but hoping not..
-		sr := spriteRender{drawPos: sp.Geom.Pos.Add(scpos), pixels: sp.Pixels, active: sp.Active}
-		ss.sprites[i] = sr
+func SpritesSource(stage *Stage, mainScene *Scene) composer.Source {
+	stage.Sprites.Lock()
+	defer stage.Sprites.Unlock()
+
+	sz := math32.FromPoint(mainScene.SceneGeom.Size)
+	if stage.spritePainter.State == nil || stage.spritePainter.State.Size != sz {
+		stage.spritePainter = *paint.NewPainter(sz)
```
kkoreilly: Please store `spritePainter` as a pointer (same for `Scene.Painter`).

core/renderwindow.go
```diff
@@ -674,6 +685,14 @@ func (w *renderWindow) renderWindow() {
 	}
 	spriteMods := top.Sprites.IsModified()
 
+	spriteUpdateTime := SystemSettings.CursorBlinkTime
+	if spriteUpdateTime == 0 {
+		spriteUpdateTime = 500 * time.Millisecond
+	}
+	if time.Since(w.lastSpriteDraw) > spriteUpdateTime {
+		spriteMods = true
```
kkoreilly, line 694: This jeopardizes the idea of a performant app that only updates when it needs to...

kkoreilly (changes requested):

colors/image.go
rcoreilly: the entire point of this function is to get rid of pointers to things. `color.Color` is a pointer (interface). Need to get a concrete value type here.

core/enumgen.go
rcoreilly: new logic: unexported types get unexported values -- only exported does filtering.

core/render_js.go
rcoreilly: left a todo.

core/renderwindow.go
rcoreilly: There is exceptionally minimal actual code that runs if there are no updates.

core/sprite.go
```diff
@@ -144,83 +123,180 @@ func (sp *Sprite) send(typ events.Types, original ...events.Event) {
...
 	ss.modified = true
 	ss.Unlock()
 }
 
-// Delete deletes sprite by name, returning indexes where it was located.
-// All sprite images must be updated when this occurs, as indexes may have shifted.
-func (ss *Sprites) Delete(sp *Sprite) {
-	ss.Lock()
-	ss.DeleteKey(sp.Name)
+// SetModifiedLocked sets the sprite modified flag, which will
```
kkoreilly: It would be really nice if we could thing of a name for this kind of function that doesn't seem to imply that this version *does* lock it, when in fact it doesn't lock it, but should be called while locked. SetModifiedWhileLocked seems a little long... SetModifiedSync (as opposed to Async)?

core/stage.go
```diff
@@ -374,6 +382,7 @@ func (st *Stage) delete() {
 	if st.Type.isMain() && st.popups != nil {
 		st.popups.deleteAll()
 		st.Sprites.reset()
+		st.spriteRenderer = nil
```
kkoreilly, line 385: Why don't you also set the spritePainter to nil here?

kkoreilly (changes requested):

core/sprite.go
kkoreilly: We decided on `NoLock` (ex: `SetModifiedNoLock`).

core/table.go
```diff
@@ -305,7 +305,7 @@ func (tb *Table) MakeRow(p *tree.Plan, i int) {
 		valnm := fmt.Sprintf("value-%d-%s-%s", fli, itxt, reflectx.ShortTypeName(field.Type))
 		tags := field.Tag
 		if uv.Kind() == reflect.Slice || uv.Kind() == reflect.Map {
-			ni := reflect.StructTag(`display:"no-inline"`)
+			ni := reflect.StructTag(`display:"no-inline" new-window:"+"`)
```
kkoreilly, line 308: Instead of gradually making more things NewWindow by default, what about adding a user setting? You like NewWindow, but I don't really.

kkoreilly (changes requested):

core/table.go
kkoreilly: I filed #1554 for this.

svg/io.go
```diff
@@ -933,28 +938,26 @@ func MarshalXML(n tree.Node, enc *XMLEncoder, setName string) string {
...
+			XMLAddAttr(&se.Attr, "style", sp)
+		}
+		if txp, has := properties["transform"]; has {
+			XMLAddAttr(&se.Attr, "transform", reflectx.ToString(txp))
+		}
+	} else {
+		for k, v := range properties {
+			sv := reflectx.ToString(v)
+			if _, has := InkscapeProperties[k]; has {
+				k = "inkscape:" + k
```
kkoreilly, line 955: Can we just strip the inkscape properties?

core/stage.go
rcoreilly: because it wasn't a pointer before :)

kkoreilly (changes requested):

svg/io.go
rcoreilly: No we use them for metadata. we are only saving the ones we use now.

system/driver/base/window.go
```diff
@@ -82,7 +82,8 @@ func (w *Window[A]) WinLoop() {
 	if fps <= 0 {
 		fps = 60
 	}
-	winPaint := time.NewTicker(time.Second / time.Duration(fps))
+	dur := time.Second / time.Duration(fps)
+	winPaint := time.NewTicker(dur)
```
kkoreilly, line 86: So you didn't end up needing the no ticker approach, or is there still an issue with the ticker no longer firing?

text/textcore/base.go
```diff
@@ -199,7 +199,8 @@ func (ed *Base) Init() {
 	ed.Styles.Font.Family = rich.Monospace // critical
 	ed.SetLines(lines.NewLines())
 	ed.Styler(func(s *styles.Style) {
-		s.SetAbilities(true, abilities.Activatable, abilities.Focusable, abilities.Hoverable, abilities.Slideable, abilities.DoubleClickable, abilities.TripleClickable)
+		s.SetAbilities(true, abilities.Activatable, abilities.Hoverable, abilities.Slideable, abilities.DoubleClickable, abilities.TripleClickable)
+		s.SetAbilities(!ed.IsReadOnly(), abilities.Focusable)
```
kkoreilly, line 203: I thought we wanted to support focusing and cursor movement in read only text editors.

kkoreilly (changes requested):

system/driver/base/window.go
rcoreilly: correct, ticker was not the source of the problem . instead, it was solved by an atomic counter (basically a waitgroup).

text/textcore/base.go
rcoreilly: why again? seems more confusing than not to have a cursor when you can't edit.

kkoreilly: Don't you like to use keyboard navigation in Cogent Code on output read-only text editors? It is confusing to have a cursor when you can't edit, but I thought you needed it.

rcoreilly: yeah not worth the confusion.

kkoreilly (approved):
Finally all good!

PR #1555, opened 2025-06-17
htmlists: fix list handling in htmlcore 

rcoreilly:
maintains list depth, indent, proper ordering etc


rcoreilly:
@kkoreilly this is ready for review. in principle cogent/web should be working much better with the addition of tables and fixed lists, but it seems to have some other issues -- might want to fix that up and see how far we are.

htmlcore/handler.go
```diff
@@ -386,6 +400,24 @@ func handleText(ctx *Context) *core.Text {
 	return tx
 }
 
+// handleTextExclude creates a new [core.Text] from the given information, setting the text and
+// the text click function so that URLs are opened according to [Context.OpenURL].
+// excludeSubs is a list of sub-node types to exclude in processing this element.
+// If one of those types is encountered, it is returned.
+func handleTextExclude(ctx *Context, excludeSubs ...string) (*core.Text, *html.Node) {
```
kkoreilly, line 398: It would presumably be much more efficient to make HandleText call HandleTextExclude with nothing to exclude, and the same for ExtractText and ExtractTextExclude, instead of unnecessarily duplicating all of the logic.

kkoreilly (changes requested):

htmlcore/handler.go
rcoreilly: ExtractText is called a number of times separately, and the Exclude version requires an extra return value, which is a pita to deal with in the regular calls. But yeah I was able to get rid of one function.

kkoreilly: Thanks for doing that. Why not also do it for handleText as I asked? `handleText` can just call `handleTextExclude`.

rcoreilly: same reason: handleText has 1 return value and it is used inline in many cases.

rcoreilly: ok handleText now calls handleTextExclude

htmlcore/text.go
```diff
@@ -21,31 +21,63 @@ func ExtractText(ctx *Context) string {
...
 	if !it {
+		sub := n.Data
+		excl := false
+		for _, ex := range excludeSubs {
+			if ex == sub {
+				excl = true
+				break
+			}
+		}
+		if excl {
```
kkoreilly: Replace this and the for loop and vars above with a simple `if slices.Contains(excludeSubs, n.Data) {`.

kkoreilly (changes requested):

htmlcore/text.go
rcoreilly: yeah I need to remember that one.. eventually.. :)

kkoreilly (approved):

PR #1558, opened 2025-06-21
Content supports stdout printing into a texteditor.

rcoreilly:
verified that it works on the web too.

### DIFF
--- a/base/exec/stdio.go
+++ b/base/exec/stdio.go
@@ -49,7 +49,7 @@ func (st *StdIO) Set(o *StdIO) *StdIO {
 // SetToOS sets the current IO to os.Std*, returning
 // a StdIO with the current IO settings prior to this call,
 // which can be used to restore.
-// Note: os.Std* are *os.File types, and this function will panic
+// Note: os.Std* are *os.File types, and this function will do nothing
 // if the current IO are not actually *os.Files.
 // The results of a prior SetToOS call will do the right thing for
 // saving and restoring the os state.
--- a/base/exec/stdio.go
+++ b/base/exec/stdio.go
@@ -61,8 +61,16 @@ func (st *StdIO) SetToOS() *StdIO {
 	} else {
 		fmt.Printf("In is not an *os.File: %#v\n", st.In)
 	}
-	os.Stdout = st.Out.(*os.File)
-	os.Stderr = st.Err.(*os.File)
+	if st.Out != nil {
+		if so, ok := st.Out.(*os.File); ok {
+			os.Stdout = so
+		}
+	}
+	if st.Err != nil {
+		if so, ok := st.Err.(*os.File); ok {
+			os.Stderr = so
+		}
+	}
 	return cur
 }
 
--- a/text/lines/layout.go
+++ b/text/lines/layout.go
@@ -130,6 +130,9 @@ func (ls *Lines) layoutViewLine(ln, width int, txt []rune, mu rich.Text) ([]rich
 	muls := make([]rich.Text, 0, nb+1)
 	last := 0
 	for _, si := range breaks {
+		if si < last {
+			continue
+		}
 		muls = append(muls, lt[last:si])
 		last = si
 	}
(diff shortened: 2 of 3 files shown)


base/exec/stdio.go
```diff
@@ -61,8 +61,16 @@ func (st *StdIO) SetToOS() *StdIO {
 	} else {
 		fmt.Printf("In is not an *os.File: %#v\n", st.In)
 	}
-	os.Stdout = st.Out.(*os.File)
-	os.Stderr = st.Err.(*os.File)
+	if st.Out != nil {
+		if so, ok := st.Out.(*os.File); ok {
```
kkoreilly, line 65: When is this check necessary?

text/lines/layout.go
```diff
@@ -98,7 +99,11 @@ func (ls *Lines) layoutViewLine(ln, width int, txt []rune, mu rich.Text) ([]rich
 					lt.SplitSpan(i)
 					si++
 				}
-				breaks = append(breaks, si)
+				if si < 0 {
+					fmt.Println("lines.layout si < 0")
```
kkoreilly, line 103: Does this happen? How about a clearer log message so this doesn't become like the old `aberrant sp`? Same with the one below.

yaegicore/yaegicore.go
```diff
@@ -16,12 +17,17 @@ import (
...
 	"cogentcore.org/core/htmlcore"
+	"cogentcore.org/core/styles"
 	"cogentcore.org/core/text/textcore"
 	"cogentcore.org/core/yaegicore/basesymbols"
 	"cogentcore.org/core/yaegicore/coresymbols"
 	"github.com/cogentcore/yaegi/interp"
 )
 
+var (
+	InterpOutput bytes.Buffer
```
kkoreilly: Why a global var?

kkoreilly (changes requested):

base/exec/stdio.go
rcoreilly: because it crashes otherwise..

text/lines/layout.go
rcoreilly: yeah it does happen -- 2x in past week in my Code crash log. hoping to figure out why. This is all just for my personal consumption -- nobody else should see it!

yaegicore/yaegicore.go
rcoreilly: original impl was using this in yaegilab -- still might want to have it avail? could wait until that is necc. I guess.

base/exec/stdio.go
rcoreilly: I don't think we actually need / want to be updating os.Std* guys in the first place, so I will likely remove this once I go through everything and make this much more robust at some point, in coordination with Goal.

yaegicore/yaegicore.go
kkoreilly: I asked why it was a global var, not why it was exported. Shouldn't it just be in the function?

rcoreilly: no because you can see in the code that it is created in one place and then needs to be used in another.

kkoreilly: It is created in `getInterpreter`, which is called by `BindTextEditor`, which also uses it, so it can and should be made in BindTextEditor. Otherwise it is a potential race condition; each text editor should get a separate place for storing output!

rcoreilly: but BindTextEditor doesn't know if it is new or old...

rcoreilly: and can't get the stdout from interpreter.

rcoreilly: how about you rewrite this one -- I don't see a good soln.

kkoreilly: Okay, I see now that the problem is with the goal interpreter. It would work to have it as a local variable for non-goal interpreters, but that isn't worth it.

kkoreilly (approved):

PR #1567, opened 2025-07-05
keymaps: add keymaps docs to Resources category.

rcoreilly:
also fix htmlcore table to properly deal with empty cells.


docs/content/keymaps.md
```diff
@@ -0,0 +1,285 @@
++++
+Categories = ["Resources"]
++++
+
+The [[doc:keymap]] package maps keyboard input [[events]] into a [[doc:keymap.Functions]] enum, which has a standard list of app functions that are typically activated by keyboard events. [[Settings]] allows the user to select from a set of different standard keyboard mappings, which are documented in the tables on this page.
```
kkoreilly: Bold keymap.

kkoreilly: And Settings *allow*, not allows.

yaegicore/yaegicore.go
```diff
@@ -49,27 +50,36 @@ type Interpreter interface {
 	Eval(src string) (res reflect.Value, err error)
 }
 
-var autoPlanNameCounter uint64
-
 func init() {
 	htmlcore.BindTextEditor = BindTextEditor
 	coresymbols.Symbols["."] = map[string]reflect.Value{} // make "." available for use
 	basesymbols.Symbols["."] = map[string]reflect.Value{} // make "." available for use
+	content.NewPageInitFunc = ResetGoalInterpreter
```
kkoreilly: Move this up to be right below `htmlcore.BindTextEditor`, which is a very similar approach.

kkoreilly (changes requested):

docs/content/keymaps.md
rcoreilly: you don't bold a link, right? it is already highlighted!

kkoreilly: The subject of each page should be bolded in the first instance in the intro paragraph, per wikipedia style. It shouldn't be a link in general, but it if is, you should still bold it.

kkoreilly (approved):

PR #1578, opened 2025-07-13
Imgresize: do resize and cropping on the web, handled in the styles.ObjectFit code.

rcoreilly:
should make web content rendering with images much faster.


base/iox/imagex/wrapjs_js.go
```diff
@@ -45,20 +91,38 @@ type JSRGBA struct {
...
 		return im
+	case *image.RGBA:
+		im.RGBA = x
+		im.JS.SetImageData(x.Pix, src.Bounds(), options)
+	case *image.NRGBA:
+		if options == nil {
+			options = map[string]any{"premultiplyAlpha": "premultiply"}
+		} else {
+			options["premultiplyAlpha"] = "premultiply"
+		}
```
kkoreilly:
```go
		if options == nil {
			options = map[string]any{}
		}
		options["premultiplyAlpha"] = "premultiply"
```

core/renderwindow.go
```diff
@@ -236,6 +236,20 @@ func (w *renderWindow) logicalDPI() float32 {
...
+// in certain rare cases.
+func (w *renderWindow) devicePixelRatio() float32 {
+	if w.SystemWindow == nil || w.SystemWindow.Screen() == nil {
+		sc := system.TheApp.Screen(0)
+		if sc == nil {
+			return 1 // default
+		}
+		return sc.DevicePixelRatio
+	}
+	return w.SystemWindow.Screen().DevicePixelRatio
```
kkoreilly: This code is excessively nested; I will fix.

kkoreilly (changes requested):
Image rendering is entirely not working on js for me with this PR...

core/renderwindow.go
kkoreilly: No one even calls this function...

rcoreilly: I did in the course of diagnosing the issue. It is a small and reasonable function to provide.

kkoreilly: It is unexported. An unexported and unused function is not really helpful.

kkoreilly (approved):

PR #1580, opened 2025-07-14
nestplus: restore nest:+ functionality to cli

rcoreilly:
needed for emergent config to avoid cluttering top-level space with items that are semantically only meaningful in their full form.  e.g., Params.SaveAll is only about saving params, not about anything else, so just doing -save-all is just not right..


cli/field.go
```diff
@@ -194,6 +194,10 @@ func addFieldsImpl(obj any, path string, cmdPath string, allFields *fields, used
 				}
 			} else {
 				// if no conflict, we get the name
+				nfns := nf.Field.Tag.Get("nest")
+				if nfns == "+" || nfns == "true" {
+					applyShortestUniqueName(nf, i, usedNames)
```
kkoreilly: `applyShortestUniqueName` does exactly what the name describes. If there are no conflicts, it will apply the shortest name possible, which is just the name of the field. This does not accomplish what you probably intend. You should probably test the sims and see what the behavior is and whether it is actually problematic, and if it is, make sure that your changes fix the issue.

kkoreilly (changes requested):

cli/field.go
rcoreilly: as discussed, you will make a more explicit function that does the right thing. this actually does work because it conflicts with itself or something like that.

kkoreilly: It turns out that a bizarre combination of incorrect behavior was causing this to work. For example, with `Run.Debug` with `nest:"+"`, `shortestUniqueName` was returning Debug as it should, but that was causing `applyShortestUniqueName` to *delete* Debug, since the index you are passing corresponds to Debug, when it should correspond to Run.Debug, but this is the wrong context for that.

This led me to discover a similar issue taking place with something like `Run.Run`, where Run would end up without a name because, in the conflict resolution, it would get Run as the shortest name, but that was the same one it already had, so it would delete it because it thought it was a duplicate. I fixed this issue with a key check in `applyShortestUniqueName`, which also led your code to correctly not work.

I then implemented the correct solution for `nest:"+"`, so this is all working now!

kkoreilly (approved):

PR #1581, opened 2025-07-14
pw: remove intermediate parent-relative sizing calls from layout

### LINKED ISSUE #1516: Split Content Spacing Issue
### Describe the bug

While creating a UI layout for large content in the center of a split I ran in to an issue where the scrollbars get doubled.  It looks like the split's "handles" maybe aren't accounted for?  The outer scrollbar is very small but very confusing, it can be fixed by setting the overflow of the center to be hidden, but I worry that I may be masking some other issue?

### How to reproduce

1. Run example code
2. Scroll in center to see two scroll bars for outer and inner frames
3. Un-comment line 31 to hide overflow in the "centerFrame"
4. Re-run code, only one set of scroll bars, works nice

### Example code

```go
package main

import (
	"cogentcore.org/core/core"
	"cogentcore.org/core/styles"
)

func main() {
	b := core.NewBody("Split Scrollbar Bug?")

	splitFrame := core.NewSplits(b)
	splitFrame.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.Direction = styles.Row
	})
	splitFrame.SetSplits(20, 60, 20)

	firstFrame := core.NewFrame(splitFrame)
	firstFrame.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.Border.Width.Set(units.Dp(4))
		s.CenterAll()
	})
	core.NewText(firstFrame).SetText("20% Split Frame")

	// The center of the split, created so that we can put child widgets in it
	centerFrame := core.NewFrame(splitFrame)
	centerFrame.Styler(func(s *styles.Style) {
		s.Direction = styles.Column
		//s.Overflow.Set(styles.OverflowHidden)
	})

	// represents a top "menu" bar that has buttons in it
	centerMenu := core.NewFrame(centerFrame)
	centerMenu.Styler(func(s *styles.Style) {
		s.Min.Set(units.Pw(100), units.Ph(10))
		s.Max.Set(units.Pw(100), units.Ph(10))
		s.Border.Width.Set(units.Dp(4))
		s.CenterAll()
	})
	core.NewText(centerMenu).SetText("Menu Bar")

	// the frame that the child widget will use to wrap its content so that from the outside we only have a single frame
	centerContentArea := core.NewFrame(centerFrame)
	centerContentArea.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.Border.Width.Set(units.Dp(4))
		s.Overflow.Set(styles.OverflowScroll)
	})

	// represents content that will be too large for this frame, user will need to scroll this on both
	// axes to interact with it correctly
	internalCenter := core.NewFrame(centerContentArea)
	internalCenter.Styler(func(s *styles.Style) {
		s.Min.Set(units.Pw(125), units.Ph(125))
		s.CenterAll()
	})
	core.NewText(internalCenter).SetText("Content in center frame")

	lastFrame := core.NewFrame(splitFrame)
	lastFrame.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.CenterAll()
		s.Border.Width.Set(units.Dp(4))
	})
	core.NewText(lastFrame).SetText("20% Split Frame")

	b.RunMainWindow()
}
```

### Relevant output

```shell

```

### Platform

Linux

### LINKED ISSUE #1517: Parent-based sizing (Pw and Ph) can be unstable
### Describe the bug

When specifying sizing in terms of Pw or Ph (especially Pw), the size of the widget you are styling can often affect the size of the parent, leading to recursion or instability.

### How to reproduce

See #1516 for an example. This also can happen with image sizing in htmlcore. More generally, having Pw and Ph correspond to *visible* parent width and height would potentially prevent this problem. See also #1518 for another case where visible parent size should be used.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

macOS

rcoreilly:
seems to work. fixes #1516 fixes #1517

to be clear: I tried the example in #1516 and it worked as expected. the inner splitter was always 125%, and there was no outer scrollbar.

### DIFF
--- a/core/layout.go
+++ b/core/layout.go
@@ -650,17 +650,18 @@ func (wb *WidgetBase) updateParentRelSizes() bool {
 		return false
 	}
 	sz := &wb.Geom.Size
+	effmin := sz.Min
 	s := &wb.Styles
 	psz := pwb.Geom.Size.Alloc.Content.Sub(pwb.Geom.Size.InnerSpace)
 	got := false
 	for d := math32.X; d <= math32.Y; d++ {
 		if s.Min.Dim(d).Unit == units.UnitPw {
 			got = true
-			sz.Min.SetDim(d, psz.X*0.01*s.Min.Dim(d).Value)
+			effmin.SetDim(d, psz.X*0.01*s.Min.Dim(d).Value)
 		}
 		if s.Min.Dim(d).Unit == units.UnitPh {
 			got = true
-			sz.Min.SetDim(d, psz.Y*0.01*s.Min.Dim(d).Value)
+			effmin.SetDim(d, psz.Y*0.01*s.Min.Dim(d).Value)
 		}
 		if s.Max.Dim(d).Unit == units.UnitPw {
 			got = true
@@ -672,8 +673,8 @@ func (wb *WidgetBase) updateParentRelSizes() bool {
 		}
 	}
 	if got {
-		sz.FitSizeMax(&sz.Actual.Total, sz.Min)
-		sz.FitSizeMax(&sz.Alloc.Total, sz.Min)
+		sz.FitSizeMax(&sz.Actual.Total, effmin)
+		sz.FitSizeMax(&sz.Alloc.Total, effmin)
 		sz.setContentFromTotal(&sz.Actual)
 		sz.setContentFromTotal(&sz.Alloc)
 	}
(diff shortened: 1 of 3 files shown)


rcoreilly:
ps. confirmed that this does not change any tests relative to current main, while fixing 1516

kkoreilly (approved):

PR #1585, opened 2025-07-15
maxgrow: fixed logic for grow with max on some elements; 

### LINKED ISSUE #1557: When Using Grow on Two Widgets, Defining a Max X on One Breaks Grow Behavior
### Describe the bug

If you have a parent frame that is told to grow on both axes, it will fill the screen as expected.
If you place two widgets inside of that parent frame, and tell the both to grow, they will both consume 50% of the frame as expected.
If you set a max x dimension on one of the frames, the other frame will not grow to fill the remainder of the space.  This is true regardless of which of the two frames you set the max on.

### How to reproduce

Run example code, observe two widgets sharing the parent frame space
Un-comment one or the other max x, observe the non constrained widget still consuming roughly 50% of the space, and the other consuming space up to its max

### Example code

```go
package main

import (
	"cogentcore.org/core/core"
	"cogentcore.org/core/styles"
	"cogentcore.org/core/styles/abilities"
)

func main() {
	b := core.NewBody()

	container := core.NewFrame(b)
	container.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.Border.Width.Set(units.Dp(4))
	})

	firstPane := core.NewFrame(container)
	firstPane.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.SetAbilities(true, abilities.Clickable)
		s.Border.Width.Set(units.Dp(2))
		//s.Max.X.Dp(400)
	})

	secondPane := core.NewFrame(container)
	secondPane.Styler(func(s *styles.Style) {
		s.Grow.Set(1, 1)
		s.Border.Width.Set(units.Dp(2))
		//s.Max.X.Dp(400)
	})

	b.RunMainWindow()
}
```

### Relevant output

```shell

```

### Platform

Linux

rcoreilly:
allocate leftover to others; fixes #1557


core/layout_test.go
```diff
@@ -272,3 +272,139 @@ func TestCustomLayoutButton(t *testing.T) {
 	})
 	b.AssertRender(t, "layout/custom-button")
 }
+
+func TestLayoutMaxGrow(t *testing.T) { // issue #1557
```
kkoreilly, line 342: You could use a for loop to implement all of these tests instead of having separate functions..

kkoreilly (changes requested):

core/layout_test.go
rcoreilly: how about you contribute that part :)

kkoreilly (approved):

PR #1593, opened 2025-07-16
Move core command to root for easier installation

kkoreilly:
This PR moves the `core` command to the root of the repository for easier installation. Installing now requires this command, which the docs have been updated to reflect:

```
go install cogentcore.org/core@main
```

This PR also reorganizes the structure of the cmd packages to reduce unnecessary nesting of folders.


rcoreilly (approved):
yay, finally!

PR #1595, opened 2025-07-17
laynodebug: don't print debug message unless layout debug flag set

rcoreilly:
getting this transiently during layout now with the new pw and max changes -- final layout looks good still. also got a crash due to visibility change in toolbar -- always need to protect cell access, and increase robustness by doing continue instead of break which could leave stuff unallocated. will test extensively on code


rcoreilly:
seems all good -- no issues noted across code, canvas, and sims..

kkoreilly (approved):

PR #1605, opened 2025-07-30
fieldvalue: FieldWidgeter interface for FieldWidget method for custom widgets for fields

### LINKED ISSUE #1204: Support form and table field widget customization with FieldWidget
### Describe the feature

I'd like a string field in a form struct to be rendered as a `Chooser`.

I've had a quick dig through the code and it doesn't seem possible as things stand. Essentially this equates to a larger feature request, that form field widgets can be specified via struct tags where appropriate (and in this case, enum values provided).

Code below is indicitive of what I'd like to be possible.

### Relevant code

```go
type MyStruct struct {
	SomeField string `widget:"chooser" values:"foo,bar,baz"`
}

form := core.NewForm(someFrame)
form.SetStruct(&MyStruct{}) // SomeField is rendered as a Chooser with available values of foo, bar, baz
```

rcoreilly:
Fixes #1204 

Still need to do Table -- Form is working

### DIFF
--- a/core/table.go
+++ b/core/table.go
@@ -292,12 +292,7 @@ func (tb *Table) MakeRow(p *tree.Plan, i int) {
 	si, _, invis := svi.SliceIndex(i)
 	itxt := strconv.Itoa(i)
 	val := tb.sliceElementValue(si)
-	var valPtr any
-	if si < tb.SliceSize {
-		valPtr = reflectx.UnderlyingPointer(tb.sliceUnderlying.Index(si)).Interface()
-	} else {
-		valPtr = reflectx.UnderlyingPointer(reflectx.SliceElementValue(tb.Slice)).Interface()
-	}
+	valPtr := reflectx.UnderlyingPointer(val).Interface()
 
 	if tb.ShowIndexes {
 		tb.MakeGridIndex(p, i, si, itxt, invis)
--- a/core/table.go
+++ b/core/table.go
@@ -292,7 +292,12 @@ func (tb *Table) MakeRow(p *tree.Plan, i int) {
 	si, _, invis := svi.SliceIndex(i)
 	itxt := strconv.Itoa(i)
 	val := tb.sliceElementValue(si)
-	// stru := val.Interface()
+	var valPtr any
+	if si < tb.SliceSize {
+		valPtr = reflectx.UnderlyingPointer(tb.sliceUnderlying.Index(si)).Interface()
+	} else {
+		valPtr = reflectx.UnderlyingPointer(reflectx.SliceElementValue(tb.Slice)).Interface()
+	}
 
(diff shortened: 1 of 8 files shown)


core/table.go
```diff
@@ -292,7 +292,12 @@ func (tb *Table) MakeRow(p *tree.Plan, i int) {
 	si, _, invis := svi.SliceIndex(i)
 	itxt := strconv.Itoa(i)
 	val := tb.sliceElementValue(si)
-	// stru := val.Interface()
+	var valPtr any
+	if si < tb.SliceSize {
+		valPtr = reflectx.UnderlyingPointer(tb.sliceUnderlying.Index(si)).Interface()
+	} else {
+		valPtr = reflectx.UnderlyingPointer(reflectx.SliceElementValue(tb.Slice)).Interface()
+	}
```
kkoreilly: I guess I don't understand why you can't just do `valPtr := reflectx.UnderlyingPointer(val).Interface()`. Your logic is replicating what sliceElementValue does, and then you can just get the pointer of it.

core/valuer.go
```diff
@@ -155,6 +155,34 @@ func toValue(value any, tags reflect.StructTag) Value {
 	return NewTextField() // final fallback
 }
 
+// FieldValuer is an interface that struct types can implement to specify the
+// [Value] that should be used to represent specific fields in the GUI,
+// via the FieldWidget method. For Form and Table widgets.
+type FieldValuer interface {
```
kkoreilly: This is now `FieldWidgeter`, right?

core/valuer.go
```diff
@@ -155,6 +155,34 @@ func toValue(value any, tags reflect.StructTag) Value {
...
+}
+
+// NewFieldValue converts the given value into an appropriate [Value]
+// whose associated value is bound to the given value. The given value must
+// be a pointer. It uses the given optional struct tags for additional context
+// and to determine styling properties via [styleFromTags]. It also adds the
+// resulting [Value] to the given optional parent if it specified. The specifics
+// on how it determines what type of [Value] to make are further
+// documented on [toValue].
+func NewFieldValue(field string, stru any, value any, tags reflect.StructTag, parent ...tree.Node) Value {
```
kkoreilly: Need to update docs (I will do).

core/valuer.go
kkoreilly, line 1: `Valuer` has a method `Value`, so I don't know where we got the whole `Core` prefix idea... Value is a pretty common name, so it might be good to rename that to Widget as well, although Widgeter doesn't really convey the meaning of the interface... We do implement this interface sometimes, but never in a context that is strictly necessary. We should discuss this.

kkoreilly (changes requested):

kkoreilly (approved):

PR #1609, opened 2025-08-09
wgpu: update to v25 of wgpu_native libraries

rcoreilly:
This updates core/gpu to the API changes in https://github.com/cogentcore/webgpu/pull/8 -- it won't work until that is merged and the libraries are updated and committed.  It can be used until then by following the directions on that PR.


rcoreilly:
We are going to switch over to using oliverbestmann/webgpu instead of our version. See #1676 -- closing this one.

PR #1612, opened 2025-08-20
contimage: content image handling updates: scrollbars, double-click

rcoreilly:


rcoreilly:
This is now done more cleanly and only for content, in #1614

PR #1613, opened 2025-08-26
Export Tabs.deleteTabIndex

theclapp:
Export Tabs.deleteTabIndex so you can close a tab programmatically, e.g. via ^W.


kkoreilly (approved):
Thanks for filing this and apologies for the delay in merging!

PR #1614, opened 2025-09-09
contstyle: content styling supported by new htmlcore WidgetHandlers

rcoreilly:
keeps all text elements at a max of 120ch by default for content -- looks pretty reasonable.

### DIFF
--- a/htmlcore/context.go
+++ b/htmlcore/context.go
@@ -87,7 +87,7 @@ type Context struct {
 	AttributeHandlers map[string]func(ctx *Context, w io.Writer, node ast.Node, entering bool, tag, value string) bool
 
 	// WidgetHandlers is a list of handler functions for each Widget,
-	// called in reverse order for each widget that is created,
+	// called in sequential order for each widget that is created,
 	// after it has been configured by the existing handlers etc.
 	// This allows for additional styling to be applied based on the
 	// type of widget, for example.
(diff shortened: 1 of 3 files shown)


htmlcore/context.go
```diff
@@ -244,3 +252,16 @@ func (c *Context) LinkButtonUpdating(bt *core.Button, url func() string) *core.B
...
+
+// AddWidgetHandler adds given widget handler function
+func (c *Context) AddWidgetHandler(f func(w core.Widget)) {
+	c.WidgetHandlers = append(c.WidgetHandlers, f)
+}
+
+// handleWidget calls WidgetHandlers functions on given widget.
+func (c *Context) handleWidget(w core.Widget) {
+	n := len(c.WidgetHandlers)
+	for i := n - 1; i >= 0; i-- {
```
kkoreilly: Why reverse order? Shouldn't the last added one take precedent?

kkoreilly (changes requested):

kkoreilly:
Overall looks great, although it would be more standard to center the text.

htmlcore/context.go
rcoreilly: doh -- you're right!  I outthought myself :) will fix.

kkoreilly (approved):
LGTM. Note: it was not easily possible to center the text correctly, and it doesn't seem necessary for now.

PR #1615, opened 2025-09-10
xyzplan: fully support tree Plan logic in xyz

rcoreilly:


core/scene.go
```diff
@@ -309,7 +309,7 @@ func (sc *Scene) resize(geom math32.Geom2DInt) bool {
 	sc.renderer.SetSize(units.UnitDot, sz)
 	sc.SceneGeom.Size = geom.Size // make sure
 
-	sc.updateScene()
+	// sc.updateScene() // note: doing this every resize can be very expensive, and is unnec in general. really you just need a layout.
```
kkoreilly, line 312: Given that resizing is the most convenient way to force an update, is there any alternative way to accomplish this now? Have you tested this change on web and mobile?

xyz/xyzcore/scene.go
```diff
@@ -38,9 +41,28 @@ type Scene struct {
 	SelectionParams SelectionParams `display:"inline"`
 }
 
+// NewSceneForScene returns a new [Scene] for existing [xyz.Scene],
+// in given parent (if non-nil).
+func NewSceneForScene(parent tree.Node, sc *xyz.Scene) *Scene {
```
kkoreilly: Standard practice for optional parent is var args as in typical New functions (e.g., `sc *xyz.Scene, parent ...tree.Node`). Same applies for NewSceneEditorForScene.

xyz/lines.go
```diff
@@ -135,56 +122,75 @@ const (
...
-func NewArrow(sc *Scene, parent tree.Node, name string, st, ed math32.Vector3, width float32, clr color.RGBA, startArrow, endArrow bool, arrowSize, arrowWidth float32, arrowSegs int) *Group {
-	cm := UnitConeMesh(sc, arrowSegs)
-	gp := NewGroup(parent)
-	gp.isLinear = true
-	gp.SetName(name)
+// InitLine is an Init function (e.g., for [tree] Maker functions)
+// for making a line between two specified points, using a shared
+// mesh unit line, which is rotated and positioned
+// to go between the designated points.
+func InitLine(ln *Solid, st, ed math32.Vector3, width float32, clr color.RGBA) {
```
kkoreilly, line 129: We are okay with the breaking change of removing the old New functions and only having Init functions? Do we still support imperative xyz, or does everything need to be declarative? Is this documented somewhere?

kkoreilly (changes requested):
Overall looks very good! Just a couple small comments.

core/scene.go
rcoreilly: Yeah rebuild calls it. So changing settings for example. And just confirmed works on ios, and have been using on web no prob.

xyz/xyzcore/scene.go
rcoreilly: ok fixed.

xyz/lines.go
rcoreilly: You can always impeaitively create a Solid and then call this function. We could add a wrapper to do so that replicates the prior api -- you want that?

kkoreilly: Probably don't need wrappers unless it becomes an issue.

kkoreilly (approved):

PR #1617, opened 2025-09-25
scroll: fix scrolling on mobile, vs. selection: gets stuck in text selection mode and you can't scroll anymore!

rcoreilly:


kkoreilly (approved):
This seems like a reasonable workaround, although the general thing that we want to implement at some point is to only enable text selection mode on long press. I am fine with merging this for now though, assuming you have tested properly on mobile.

PR #1619, opened 2025-10-02
PDF rendering support

rcoreilly:
converting code from tdewolff canvas PDF rendering to support direct rendering to PDF files.


colors/matcolor/scheme.go
```diff
@@ -117,14 +117,14 @@ func NewLightScheme(p *Palette) Scheme {
 		Custom:    map[string]Accent{},
 
 		SurfaceDim:    p.Neutral.AbsToneUniform(87),
-		Surface:       p.Neutral.AbsToneUniform(98),
-		SurfaceBright: p.Neutral.AbsToneUniform(98),
+		Surface:       p.Neutral.AbsToneUniform(99),
+		SurfaceBright: p.Neutral.AbsToneUniform(99),
```
kkoreilly: These are major changes to the scheme that are not necessarily justified. Do you want different settings for GUI rendering and PDF exporting?

colors/matcolor/scheme.go
```diff
@@ -134,7 +134,7 @@ func NewLightScheme(p *Palette) Scheme {
 		InverseOnSurface: p.Neutral.AbsToneUniform(95),
 		InversePrimary:   p.Primary.AbsToneUniform(80),
 
-		Background:   p.Neutral.AbsToneUniform(98),
+		Background:   p.Neutral.AbsToneUniform(100),
```
kkoreilly: Background and Surface have been the same, and we planned to eventually get rid of Background due to its redundancy. It is not clear that we should make them different.

content/bcontent/page.go
```diff
@@ -52,8 +52,16 @@ type Page struct {
 	// DateString is only used for parsing the date from the TOML front matter.
 	DateString string `toml:"Date" json:"-"`
 
-	// Authors are the optional authors of the page.
-	Authors []string
+	// Authors are the optional author(s) of the page.
+	Authors string
```
kkoreilly, line 56: Why a string not a slice?

content/buttons.go
```diff
@@ -74,6 +74,12 @@ func (ct *Content) MakeToolbar(p *tree.Plan) {
 			ct.Scene.MenuSearchDialog("Search", "Search "+core.TheApp.Name())
 		})
 	})
+	tree.Add(p, func(w *core.Button) {
+		w.SetText("PDF").SetIcon(icons.PictureAsPdf).SetTooltip("PDF generates and opens / downloads the current page as a printable PDF file. See the Settings/Printer panel (Command+,) for settings.")
```
kkoreilly, line 84: Placing this in the main toolbar by default is rather aggressive.

content/content.go
```diff
@@ -126,6 +129,7 @@ func (ct *Content) Init() {
 	ct.SetSplits(0.2, 0.8)
 
 	ct.Context = htmlcore.NewContext()
+	ct.Context.DelayedImageLoad = false // not useful for content
```
kkoreilly, line 132: Why?

content/content.go
```diff
@@ -224,19 +164,19 @@ func (ct *Content) Init() {
 				}
 			})
 			w.Maker(func(p *tree.Plan) {
-				if ct.currentPage.Title != "" {
+				if !ct.inPDFRender && ct.currentPage.Title != "" {
```
kkoreilly, line 167: Are the title and authors not valuable in the PDF rendering as well?

content/content.go
```diff
@@ -529,3 +477,71 @@ func (ct *Content) setStageTitle() {
...
+
+	ct.inPDFRender = true
+	ct.reloadPage()
+	ct.inPDFRender = false
+
+	refs := ct.PageRefs(ct.currentPage)
+
+	fname := ct.currentPage.Name + ".pdf"
+	if path != "" {
+		os.MkdirAll(path, 0777)
```
kkoreilly: errors.Log

content/content.go
```diff
@@ -529,3 +477,71 @@ func (ct *Content) setStageTitle() {
...
+	} else {
+		paginate.PDF(f, opts.PDF, ct.rightFrame)
+	}
+	err = f.Close()
+
+	ct.reloadPage()
+
+	core.MessageSnackbar(ct, "PDF saved to: "+fname)
+	af := errors.Log1(filepath.Abs(fname))
+	core.TheApp.OpenURL("file://" + af)
```
kkoreilly, line 518: What happens when you do this on web?

content/url_js.go
```diff
@@ -14,11 +14,22 @@ import (
...
+	documentData = js.Global().Get("document").Get("documentElement").Get("dataset")
+
+	// OfflineURL is the non-web base url, which can be set to allow
+	// docs to refer to this in frontmatter.
+	OfflineURL = ""
+)
+
+func (ct *Content) getPrintURL() string {
+	p := originalBase.String()
+	return p[:len(p)-1]
```
kkoreilly: Is this to trim the trailing slash? If so, strings.Trim is safer as in getWebURL below

kkoreilly (changes requested):
Comments through content.

colors/matcolor/scheme.go
rcoreilly: we need to discuss. you should pull up side-by-side and see what you think. I find it to be definitively better all around, but we could introduce an option. the prior behavior just looks "dingy" and notably out-of-sync with all other mac apps.

colors/matcolor/scheme.go
rcoreilly: k wasn't sure about that. maybe time to nuke?

content/bcontent/page.go
rcoreilly: Because it is awkward and unnecessary to have to parse and format. Just put everything as it should be in a string.

content/buttons.go
rcoreilly: ?? this is essential functionality, that works in all cases!

content/content.go
rcoreilly: content almost always has embedded images, so fully useless (and distracting) in that case. Generic HTML you're loading from the web so it is useful there.

content/content.go
rcoreilly: PDF generates its own formatted title with authors etc. so we exclude this redundant one.

content/content.go
rcoreilly: try it! amazingly you wrote something that causes it to be downloaded :)

colors/matcolor/scheme.go
kkoreilly: Yep, I would say remove it and replace references to Background with Surface (same for OnBackground).

content/buttons.go
kkoreilly: Especially having it at the start of the toolbar takes attention away from other buttons that are presumably more important for most people. I would rather have a general scene context menu button for PDF that you can use in any context (not just content), which would be quite useful.

Or if it is specific to content, still put it in a context menu, or at least not at the start of the toolbar.

content/bcontent/page.go
kkoreilly: Maybe, but there also may be some contexts where it is useful to have each author separately. For example, if in content we want to have automatically generated author pages where you can see everything that someone has written, you need each author separately.

core/settings.go
```diff
@@ -36,7 +38,17 @@ import (
...
+var AllSettings = []Settings{AppearanceSettings, SystemSettings, TimingSettings, DebugSettings}
+
+// AddAppSettings inserts app-specific settings after the AppearanceSettings
+func AddAppSettings(sets ...Settings) {
+	AllSettings = slices.Insert(AllSettings, 1, sets...)
+}
+
+// AddPrinterSettings adds printer settings, for apps that use these.
+func AddPrinterSettings() {
+	AddAppSettings(&printer.Settings)
```
kkoreilly: Does it make sense for core to always import printer? I guess the compiler can probably optimize this away if this function isn't called.

kkoreilly (changes requested):
Comments through gpu.

htmlcore/handler.go
```diff
@@ -498,8 +492,3 @@ func Get(ctx *Context, url string) (*http.Response, error) {
 	}
 	return resp, nil
 }
-
-// BindTextEditor is a function set to [cogentcore.org/core/yaegicore.BindTextEditor]
-// when importing yaegicore, which provides interactive editing functionality for Go
-// code blocks in text editors.
-var BindTextEditor func(ed *textcore.Editor, parent *core.Frame, language string)
```
kkoreilly, line 505: Why is this moved out of htmlcore to content? This is useful in htmlcore for things like live preview in Cogent Code.

htmlcore/handler.go
```diff
@@ -376,22 +295,99 @@ func handleElement(ctx *Context) {
...
+			img.AsyncUnlock()
+		}
+	}()
+	return newWidget
 }
 
 // handleText creates a new [core.Text] from the given information, setting the text and
 // the text click function so that URLs are opened according to [Context.OpenURL].
-func handleText(ctx *Context) *core.Text {
+func handleText(ctx *Context, tag string) *core.Text {
```
kkoreilly, line 390: It doesn't seem like we are using tag here?

math32/matrix2.go
```diff
@@ -39,7 +39,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-// Matrix2 is a 3x2 matrix.
+// Matrix2 is a 2x3 matrix.
```
kkoreilly: Matrix dimensions are typically MxN with M rows and N columns. According to the representation below, it looks like 3 rows and 2 columns, so why 2x3 instead of the previous 3x2?

math32/vector2i.go
```diff
@@ -17,8 +17,8 @@ type Vector2i struct {
 }
 
 // Vec2i returns a new [Vector2i] with the given x and y components.
-func Vec2i(x, y int32) Vector2i {
-	return Vector2i{X: x, Y: y}
+func Vec2i(x, y int) Vector2i {
+	return Vector2i{X: int32(x), Y: int32(y)}
```
kkoreilly, line 21: Why the inconsistent int args and int32 fields?

kkoreilly (changes requested):
Comments through math32.

math32/vector2i.go
rcoreilly: it is much more convenient in general to have plain int for this constructor.

htmlcore/handler.go
rcoreilly: true but likely future use will need. let's optimize this later.

htmlcore/handler.go
rcoreilly: The issue was that we ended up doing very content-specific things with this in content. If / when we need similar functionality in base htmlcore, for non-content, actual HTML uses, then the relevant code can be introduced here, in a way that doesn't conflict with the content use-case.

paint/renderers/renderers.go
```diff
@@ -6,10 +6,12 @@ package renderers
...
 import (
 	"cogentcore.org/core/paint"
+	"cogentcore.org/core/paint/renderers/pdfrender"
 	"cogentcore.org/core/paint/renderers/svgrender"
 	_ "cogentcore.org/core/text/shaped/shapers"
 )
 
 func init() {
 	paint.NewSVGRenderer = svgrender.New
+	paint.NewPDFRenderer = pdfrender.New
```
kkoreilly, line 16: What is the impact of always importing the pdf renderer on binary size? Should this be more optional?

If it is included in every app, then we might as well put export to pdf in the default scene context menu.

styles/style.go
```diff
@@ -248,6 +248,15 @@ func (s *Style) Defaults() {
 	s.Text.Defaults()
 }
 
+// ZeroSpace sets all the spacing elements to zero: Padding, Border, Margin, Gap
+func (s *Style) ZeroSpace() {
+	s.Padding.Zero()
+	s.Margin.Zero()
+	s.MaxBorder.Width.Zero()
+	s.Border.Width.Zero()
+	s.Gap.Zero()
```
kkoreilly, line 257: You could include BoxShadow and MaxBoxShadow here too.

system/driver/base/app.go
```diff
@@ -67,6 +69,9 @@ func Init(a system.App, ab *App) {
 	ab.This = a
 	system.TheApp = a
 	key.SystemPlatform = a.SystemPlatform().String()
+	// sl := a.SystemLocale()
+	// fmt.Println("locale:", sl, sl.Language(), sl.Region())
+	printer.Settings.Defaults() // depends on system.TheApp
```
kkoreilly: Do we have to import printer here? Could another package that imports both printer and system do this? Everything else in Init is very low level and high priority.

system/locale.go
```diff
@@ -0,0 +1,35 @@
...
+
+import "strings"
+
+// Locale represents a https://www.rfc-editor.org/rfc/bcp/bcp47.txt standard
+// language tag, consisting of language and region, e.g., "en-US", "fr-FR", "ja-JP".
+type Locale string
+
+// Language returns the language portion of the locale tag (e.g., en, fr, ja)
+func (l Locale) Language() string {
+	if l == "" {
```
kkoreilly: This whole function could be 

```go
lang, _, _ := strings.Cut(string(l), "-")
return lang
```

system/locale.go
```diff
@@ -0,0 +1,35 @@
...
+	pos := strings.LastIndex(string(l), "-")
+	if pos < 0 {
+		return string(l)
+	}
+	return string(l)[:pos]
+}
+
+// Region returns the region portion of the locale tag (e.g., US, FR, JA)
+func (l Locale) Region() string {
+	if l == "" {
```
kkoreilly: This function can be

```go
_, region, _ := strings.Cut(string(l), "-")
return region
```

text/csl/mdcite/mdcite.go
```diff
@@ -45,7 +50,24 @@ func Generate(c *Config) error {
 		return err
 	}
 	kl := csl.NewKeyList(refs)
-	cited, err := csl.GenerateMarkdown(c.Dir, c.Output, c.Heading, kl, c.Style)
+
+	if c.Dir == "" {
+		c.Dir = "./"
```
kkoreilly: Does this need to be a backslash on Windows?

kkoreilly (changes requested):
Comments up to paginate.

text/csl/mdcite/mdcite.go
kkoreilly: If so, use `filepath.Separator`.

text/paginate/paginate_test.go
```diff
@@ -0,0 +1,55 @@
...
+	// b.AssertRender(t, "text-only")
+	showed := make(chan struct{})
+	b.OnFinal(events.Show, func(e events.Event) {
+		showed <- struct{}{}
+	})
+	b.RunWindow()
+	<-showed
+
+	opts := NewOptions()
+	opts.FontFamily = rich.Serif
```
kkoreilly: opts.FontFamily undefined (type Options has no field or method FontFamily)

text/paginate/runners.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// TextStyler does standard text styling for printout:
+// FontFamily and Black Color (e.g., in case user is in Dark mode).
+func TextStyler(s *styles.Style) {
+	s.Font.Family = printer.Settings.FontFamily
+	s.Color = colors.Uniform(colors.Black)
+}
+
+// CenteredPageNumber generates a page number cenetered in the frame
```
kkoreilly: centered

text/paginate/runners.go
```diff
@@ -0,0 +1,192 @@
...
+		s.Grow.Set(1, 0)
+		s.Justify.Content = styles.Center
+	})
+	core.NewText(fr).SetText(strconv.Itoa(pageNo)).Styler(func(s *styles.Style) {
+		TextStyler(s)
+	})
+}
+
+// NoFirst excludes the first page for any runner
+func NoFirst(fun func(frame *core.Frame, opts *Options, pageNo int)) func(frame *core.Frame, opts *Options, pageNo int) {
```
kkoreilly, line 45: What about making a type for `func(frame *core.Frame, opts *Options, pageNo int)` so you don't have to repeat it in every definition?

text/rich/settings.go
```diff
@@ -11,12 +11,14 @@ import (
...
 }
 
-// DefaultSettings contains the default global text settings.
-// This will be updated from rich.DefaultSettings.
-var DefaultSettings Settings
+// Settings contains the global text settings,
+// for language, script and font names to use.
+// To use different settings temporarily, save current
+// and swap.
+var Settings SettingsData
```
kkoreilly, line 21: Global variable is better than passing every time?

kkoreilly (changes requested):
Comments through rich.

text/shaped/shaped_test.go
```diff
@@ -165,15 +162,15 @@ func TestHebrew(t *testing.T) {
...
 	})
 }
 
 func TestVertical(t *testing.T) {
-	RunTest(t, "nihongo_ttb", 300, 300, func(pc *paint.Painter, sh Shaper, tsty *text.Style, rts *rich.Settings) {
-		rts.Language = "ja"
-		rts.Script = language.Han
+	RunTest(t, "nihongo_ttb", 300, 300, func(pc *paint.Painter, sh Shaper, tsty *text.Style) {
+		// rts.Language = "ja"
+		// rts.Script = language.Han
```
kkoreilly, line 174: We don't need this anymore?

go.mod
```diff
@@ -28,6 +29,7 @@ require (
 	github.com/h2non/filetype v1.1.3
 	github.com/hack-pad/hackpadfs v0.2.1
 	github.com/jackmordaunt/icns/v2 v2.2.7
+	github.com/jeandeaual/go-locale v0.0.0-20250612000132-0ef82f21eade
```
kkoreilly, line 32: This will not work on Android because we don't call SetRunOnJVM (see https://github.com/jeandeaual/go-locale). Not a major concern, but might be good to address, or at least have more robust error handling in `System.App.SystemLocale`.

kkoreilly (changes requested):
All reviewed!

paint/renderers/renderers.go
rcoreilly: it IS already in the core namespace, via the Grab snapshot keybinding (Control+Shift+G) -- try it! drops SVG, PDF, and PNG versions!  Key point: you need a keyboard shortcut, not a menu, because otherwise the menu shows up in the snapshot!

system/driver/base/app.go
rcoreilly: IDK enough about the timing of all this stuff -- when is this Init method called? Settings are generally loaded pretty early-on.. who else will be available to know when Init is called, before settings is needed?? Not much in printer in any case -- just a couple of small files!

system/locale.go
rcoreilly: me gotta remember Cut..

text/paginate/runners.go
rcoreilly: I thought you didn't like doing that? hides the actual signature -- not a big fan.

text/rich/settings.go
rcoreilly: Absolutely. almost never need to do the swap. Not even possible in many cases to pass around. 0% chance of doing that.

text/shaped/shaped_test.go
rcoreilly: yeah it actually infers all that accurately anyway.

go.mod
rcoreilly: That sounds like a you thing. in a later PR.

paint/renderers/renderers.go
kkoreilly: Presumably we could have a menu button that uses `Defer` or something like that to wait until the menu is closed before grabbing a snapshot. It's hard for an end user to know about Control+Shift+G, and it is also a nonstandard shortcut as far as I know. What about a print button with shortcut Cmd+P?

system/driver/base/app.go
kkoreilly: Init is pretty much the very first thing that happens; other things that happen at this time include initializing the gpu etc. I guess this is fine, but it might be more consistent for core to call this. Also, if printing is supported in all apps, why does only content call AddPrinterSettings? What if we add printer settings for all apps, and set the defaults when we do that?

In fact, if every app has printer settings added in core, Defaults will automatically be called. Any Defaults method on something that implements the Settings interface is automatically called at the right time. So remove this manual calling of Defaults and add printer settings to the default core settings?

text/paginate/runners.go
kkoreilly: Yes, but in this case we have the same long signature repeated more than once per line across several lines, which looks rather ugly. When you have a function factory that transforms an existing function of the same type, a shorter name can be convenient. Up to you though, either way is probably fine.

go.mod
kkoreilly: Filed #1635.

paint/renderers/renderers.go
rcoreilly: how about another issue? sounds like it could be a bit complex and finding shortcuts is nontrivial, although the docs/keymaps page does suggest that Control+Shift+P is not taken. forget about Control+P however: that is core emacs.

content/buttons.go
rcoreilly: I'm not sure how to add a default item at the end of the toolbar. we can add it to the overflow, but not to the end. it is very small. I suggest you let this one go for now and file an issue if you really want.

text/paginate/runners.go
rcoreilly: I tried it and you have to use the sig in a few places as return values, and then it is confusing and not worth it IMO.

colors/matcolor/scheme.go
kkoreilly: I agree the new settings look pretty good. You didn't update dark mode, is that intentional?

content/bcontent/page.go
kkoreilly: Updated thoughts on this?

paint/renderers/renderers.go
kkoreilly: Do you want to try to get this working now, or should I file a separate issue? If we aren't doing this now, then what are you proposing we do with the pdf button in the toolbar?

colors/matcolor/scheme.go
rcoreilly: now fixed -- good catch!

content/bcontent/page.go
rcoreilly: I just think it is too complex to deal with the formatting issues for academic-level author strings, where for example you need to be able to put affiliation symbols and other such things. When we implement search functionality, it would be easy to search within author string for name(s) etc. If we end up needing separate author strings later, it is easier to add those and then have them format into the current Authors string, which is the definitive record, so let's just put that on the stack and deal with it if/when it becomes necessary, OK?

paint/renderers/renderers.go
rcoreilly: file separate issue. I would just keep the PDF where it is because it is small and useful and not really doing any harm, while providing key benefit for some users / sites.

colors/matcolor/scheme.go
kkoreilly: Do you want to update all of the SurfaceContainer etc colors relative to the new baseline of 0? And do you think that dark mode surface 0 even makes sense, given that other apps generally do not go fully black for dark mode (including macOS it looks like)?

rcoreilly: I'm in chrome dark mode and it looks black to me. I can't seem to find an easy way to actually decode the pixel colors but maybe you can? I think the non-0 stuff looks as "dingy" in dark mode as the non-white does in light mode. What do you think?

paint/renderers/renderers.go
kkoreilly: I moved the PDF button into a separate, optional MakeToolbarPDF function that you can call in the sites where it is useful (ccn etc). I also filed #1637.

content/buttons.go
kkoreilly: I moved the PDF button into a separate, optional MakeToolbarPDF function that you can call in the sites where it is useful (ccn etc). I also filed #1637.

colors/matcolor/scheme.go
kkoreilly: GitHub uses `rgb(13, 17, 23)` for the background (found from inspector), and macOS messages, settings, etc are clearly dark gray backgrounds, not black. Same for VS Code etc. I think we should stick with the existing for now; I reverted.

kkoreilly (approved):

PR #1640, opened 2026-01-16
Update to go v1.25.6 as standard version, and go mod updates

rcoreilly:
1.25 will soon be the older version as soon as 1.26 is released. There are various go security warnings that at least require updating to 1.24, and might as well just get ahead of the game.

* 1.25 is the only version that allows running larger wasm apps, which is critical for most of my content pages. I need to be able to run these locally to test out different browsers.
* yaegi symbol updating requires you to run a version of yaegi built with the same go version you are using. Updated our yaegi to 1.25. This is an unfortunate impediment to running different versions of go, but anyway, it makes it much easier to just have a good standard version.

This builds on imagebutton PR assuming that will be merged first. 

I tested everything extensively locally, including core run web on docs and demo. Will be updating cogent code and using that for my standard continuous testing soon.


rcoreilly:
@kkoreilly can you possibly look at the last CI run output and try to figure out why the test is saying FAIL at the end, with error code 1, when there are nothing but PASS results from the tests?

rcoreilly:
@kkoreilly it is now working.  seems so random. anyway this PR is now ready to go -- please review!

rcoreilly:
@kkoreilly I'm merging this because the updated core setup is needed for the other content repos.

Issue #1644, opened 2026-01-29
content: ctrl/cmd click opens URL in new tab; add to context menu too


rcoreilly:
### Describe the feature

Idea from Jed -- should be easy!

### Relevant code

```go

```

rcoreilly:
For non-web version, make a tabs browser and open a new tab.

rcoreilly:
fixed in #1645

Issue #1325, opened 2024-11-18
Add support for generic text searching


kkoreilly:
### Discussed in https://github.com/cogentcore/core/discussions/1317

<div type='discussions-op-text'>

<sup>Originally posted by **heejit** November 16, 2024</sup>
How do I search in document ??</div>

rcoreilly:
really need this in content!

use compile time word2vec vectors and save those for each word used in entire content space; any search word in that dict will work to find multi-word combos that match compile-time paragraph-level vectors from all the words in the param. filter on frequency etc as usual. https://en.wikipedia.org/wiki/Latent_semantic_analysis

can probably get away with short 50 dim vectors, but still that is kinda big overall.  need to do more research on other techniques.

Issue #1111, opened 2024-08-03
Docs page enhancements


suntong:
> Interactive, editable examples of all major widgets are available on the main docs page (which is also a Cogent Core app running via wasm).

Here is what issues I found about our docs site.

- Copy & paste not possible, at least with my Linux Chrome. This makes it hard for people to follow the instructions as-is, and tiny typos will make getting started to have more hurdles.
- Search is not possible, when I type something in my search box and hit enter, I'm getting this:

![image](https://github.com/user-attachments/assets/ca766458-e7f2-42f8-bf19-22edd5df9339)

Meanwhile, where is the source text for our main docs page please?

kkoreilly:
Thank you for filing this. The source for the docs is in the [docs](https://github.com/cogentcore/core/tree/main/docs) directory, mainly the [docs/content](https://github.com/cogentcore/core/tree/main/docs/content) directory. It is built using pages, which you can read about [here](https://www.cogentcore.org/core/widgets/other/pages). I am working on improving the search setup. I am also working on improving the selection support for copy and paste (see #1051).

j0hnsmith:
I noticed that double clicking a word in the docs selects the whole paragraph instead of the word, also right click doesn't display the usual menu, seems like something is overriding default os behaviours.

kkoreilly:
Yes, we are overriding default os behaviors, which is necessary for the event system to work correctly. We just need to improve text selection (#1051).

suntong:
I just noticed that middle clicking which should open the underlying link in another tab is not working either.

kkoreilly:
Yes, I will also work on fixing that.

Issue #28, opened 2018-08-20
Add support for search functionality in table and tree


rcoreilly:
In FileView, not clear if beneficial relative to completion in the textfield.. but probably some general kind of search box with < > arrows that scrolls to different rows in the tableview would be easy and generally useful.

rcoreilly:
added general search to all layouts just by typing if nothing else is in focus, and completion is probably best for fileview.  but might be good to have a demo test case of search in a table view just to work through any of those issues.

rcoreilly:
Scope: this is strictly within the widget itself.  There are other possible issues with top app bar that are related but distinct.  Still unclear of we want some kind of specific scoped search within these views.

kkoreilly:
We may want to add a global search with a text field via `Ctrl+F`.

rcoreilly:
fixed in #1645

PR #1645, opened 2026-01-30
Content fixes: ctrl+click for new tab, and fix the history

### LINKED ISSUE #1548: content back arrow behavior is incorrect
### Describe the bug

It does not take you to the immediately prior page. Something is not quite right in the stack.

### How to reproduce

1. open docs
2. click on getting started (basics page)
3. click on concepts link at top
4. click on back arrow: you go back to basics page.
5. click on build link
6. click on back arrow: you now incorrectly go back to concepts, not basics.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

macOS

rcoreilly:
This fixes #1548 #1644 #28 #1111 and mostly addresses #1325 

* full functional tabs support in content: add, delete, ctrl+click
* fixed the history logic, which is now per-tab
* Search fully supported in collection widgets (table, list, tree) and individual elements, with a Widget interface
* selection across multiple text elements

This fixes the main remaining missing features for content, and search is very long overdue.

### DIFF
--- a/content/buttons.go
+++ b/content/buttons.go
@@ -38,7 +38,7 @@ func (ct *Content) MakeToolbar(p *tree.Plan) {
 		w.SetIcon(icons.Icon(core.AppIcon))
 		w.SetTooltip("Home")
 		w.OnClick(func(e events.Event) {
-			ct.Open("")
+			ct.OpenEvent("", e)
 		})
 	})
 	// Superseded by browser navigation on web.
--- a/content/buttons.go
+++ b/content/buttons.go
@@ -47,22 +47,20 @@ func (ct *Content) MakeToolbar(p *tree.Plan) {
 			w.SetIcon(icons.ArrowBack).SetKey(keymap.HistPrev)
 			w.SetTooltip("Back")
 			w.Updater(func() {
-				w.SetEnabled(ct.historyIndex > 0)
+				w.SetEnabled(ct.historyHasBack())
 			})
 			w.OnClick(func(e events.Event) {
-				ct.historyIndex--
-				ct.open(ct.history[ct.historyIndex].URL, false) // do not add to history while navigating history
+				ct.historyBack()
 			})
 		})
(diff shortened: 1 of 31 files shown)


rcoreilly:
@kkoreilly this is ready for review. The fail is just an image that needs updated -- will do that after merge -- verified on an earlier one here.

I'll give you just a couple of days and will otherwise merge myself, and you can make suggestions or corrections later.

kkoreilly (changes requested):
Fixing the history is nice, and the cmd+click for a new tab is great (the tabs are nice overall).

There are two main issues:

1. I'm not a big fan of the new search interface. It has a lot of widgets and fails to accomplish the most important task of the current interface, which is the ability to quickly type in a page name to pull up the page. What is the motivation for the change?
2. The tabs should ideally not be rendered in html preview mode, since it is weird for them to show up in the html preview.

rcoreilly:
Thanks for the review!

1. I'll add back the original as "page search".  This new thing is a fully general command-F search that works on everything. It is huge for tables, lists, big trees, etc.
2. will fix.

rcoreilly:
ps. in a big content doc as in compcogneuro, the new search is key for finding terms etc. and even in core docs, if you are looking for a specific bit of functionality, it can get you there quickly.

rcoreilly:
@kkoreilly all good?

kkoreilly (changes requested):
I think that the toolbar should be reverted to the prior state. Text search is very valuable, but it does not need a large button at the start of the toolbar. You can make it an optional button like PDF that apps can add on their own, or you should put it in the scene context menu so that you can do it anywhere. Cmd+F is more important than the button anyone. It looks strange to have two search buttons at the top.

rcoreilly:
Ok fair point. It is already available via cmd F and the context menu as well.

rcoreilly:
@kkoreilly you're going to merge now right?

kkoreilly (approved):

PR #1663, opened 2026-03-26
glfw v3.4

rcoreilly:
Update to glfw v3.4. Currently it is just in a PR branch so this is not usable unless you go back to GOPATH mode and use a custom goal script to download all the latest packages into go/src/. 

Anyway, it all seems to work fine, so should be good when it is ready. Am now doing ongoing testing in cogent code.

TODO: update the typesetting api also here in this branch, so we can use the latest version of that (new apis for different types of glyphs -- need for emojis)


rcoreilly:
Merging this now because the new default dependencies for linux distros only take effect on CI via core@main.

PR #1668, opened 2026-04-16
Add Wayland dependencies for GLFW 3.4 on Linux

utafrali:
GLFW 3.4 added Wayland support and we need the dev libraries on Linux. Updated the package lists for all the distros we document, and regenerated the testdata while at it.


rcoreilly:
thanks!

we use a separate file system thing for the test files, so if you could delete all those testdata files, or just make a new PR with only the source changes, that would be great!

utafrali:
Done, pushed the fix.

rcoreilly:
The test files all still seem to be there -- don't see any record of a push here.

utafrali:
Done, pushed the fix.

rcoreilly:
Can't you see here that there is no record of the push, and all the files are still shown in the files changed tab?

utafrali:
Done, pushed the fix.

rcoreilly:
@utafrali I'm not sure what the issue is here but I just extracted your fixes and pushed [ad901f3](https://github.com/cogentcore/core/pull/1663/commits/ad901f3f1410156f6ec5e870d0f5567043cc2926) in #1663  -- thanks for your contribution!

PR #1672, opened 2026-04-24
Fix canvas render on web target

### LINKED ISSUE #1671: Bad canvas render on web target
### Describe the bug

On <https://www.cogentcore.org/core/canvas>, every example except `FillBox` cannot be rendered normally.

Image

### How to reproduce

See above.

### Example code

```go

```

### Relevant output

```shell

```

### Platform

Web

BioniCosmos:
Transforming the normalized coordinates using the `ctx.Cumulative` matrix will fix the issue.

## Before

Before

## After

After

Fixes #1671


rcoreilly (approved):
Thanks for finding this -- I renamed that a while back but obviously missed this case!

PR #1673, opened 2026-04-25
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `base/iox/imagex/base64.go:L43`

`FromBase64` indexes `eb[76]` without checking input length first. If attacker-controlled or malformed input shorter than 77 bytes is passed, this will panic and can be used for denial of service.

## Solution

Guard access with a length check before indexing (e.g., `if len(eb) > 76 && eb[76] == ' ' { ... }`). Also consider robust whitespace normalization that does not rely on fixed offsets.

## Changes

- `base/iox/imagex/base64.go` (modified)
- `cmd/changed.go` (modified)

### DIFF
--- a/base/iox/imagex/base64.go
+++ b/base/iox/imagex/base64.go
@@ -47,7 +47,7 @@ func Base64SplitLines(b []byte) []byte {
 
 // FromBase64 returns image from Base64-encoded bytes
 func FromBase64(eb []byte) (image.Image, Formats, error) {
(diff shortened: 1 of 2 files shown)


rcoreilly (approved):
Thanks for the fixes!

PR #1684, opened 2026-05-23
Revert "Revert "obwgpu: use oliverbestmann/webgpu instead of cogentcore/webgpu""

rcoreilly:
Reverts cogentcore/core#1678


rcoreilly:
Unfortunately the revert of the revert is the only way to move forward on this, because a merge of the old branch with main grabs the previous revert! Ugh.

PR #1704, opened 2026-06-28
xyzminor: minor fixes to xyz

rcoreilly:
also added heightfield (from waves sim) to shape, so that is avail for future reference. cleaned up a few things in plane. added tests, which only run locally because under gpu..


rcoreilly:
note: fail is one of the rare random ones.. merging

